PR #1, opened 2018-07-17
Docker init

oleksandrtarar:
Add Dockerfile and docker-compose.yml


Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM ubuntu:16.04
```
serge30: I would prefer to use `python:2.7` image, as it designed to run python code and already includes Python 2.7 and `pip`.

README.txt
```diff
@@ -0,0 +1,6 @@
+#    1. docker build -t pyramid .
```
serge30: Please use MarkDown format for documentation and file name has to be `README.md`. Add sections describing setup and info about ports available. Other info might be added during development.

Is step `docker build -t pyramid .` needed here? I believe `docker-compose up` performs it.

Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+RUN apt-get update -y && \
+    apt-get install -y python-pip python-dev && \
+    pip install --upgrade pip setuptools
+
+COPY ./requirements.txt /requirements.txt
+
+WORKDIR /app
+
+RUN pip install -r /requirements.txt
```
serge30: Please add `EXPOSE 6543` statement. It's not mandatory, but it helps to understand which ports container will be listen to.

docker-compose.yml
```diff
@@ -0,0 +1,29 @@
+version: '2'
```
serge30: Any reasons to do not use latest format version? I think version `3` might be good here.

docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+          POSTGRES_DB: eventme
+        volumes:
+          - ~/.eventme-postgres:/var/lib/postgresql/data
+    adminer:
+        image: adminer
+        restart: always
+        depends_on:
+            - db
+        ports:
+          - 9000:8080
```
serge30: Please configure you editor to add empty line to the end of file. It's common practice.

serge30 (changes requested):

Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM python:2.7
+
+RUN apt-get update -y && \
+    pip install --upgrade pip setuptools
```
serge30: As `python` image already has `pip` I think we may skip step to update deb packages, pip and setuptools.

README.md
```diff
@@ -0,0 +1,26 @@
+### Docker ###
+
+    Open your favorite Terminal and run these commands.
```
serge30: Please remove indentation, because it makes formatting incorrect. You may check MarkDown formatting using https://dillinger.io/

serge30 (changes requested):

serge30 (approved):
Looks good! You can merge it.

docker-compose.yml
oleksandrtarar: I already have 2 empty lines at the end of file, but "github" doesn't show them.

PR #7, opened 2018-07-27
Feature/log in out

Merteg:
Add two routes and two views for log-in and log-out. Add new table for tokens storage. Add seeds and migrations.


server/server/models/user.py
```diff
@@ -20,9 +23,31 @@ class User(Base):
     status_id = Column(Integer, ForeignKey("user_statuses.id"))
     role_id = Column(Integer, ForeignKey("roles.id"))
     avatar = Column(String)
+    unban_date = Column(DateTime)
```
serge30: `banned_to_date` sounds better

server/server/models/token.py
```diff
@@ -0,0 +1,28 @@
...
+    token = Column(String, unique=True)
+    user_id = Column(Integer, ForeignKey('users.id'))
+    expiration_date = Column(DateTime)
+
+    user = relationship("User", foreign_keys=(user_id,))
+
+    @classmethod
+    def add_token(cls, req, token, user_id):
+        token_obj = cls(token=token, user_id=user_id,
+                        expiration_date=datetime.now()+timedelta(days=7))
```
serge30: [Magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) is an evil :) Please extract `7` to constant with meaningful name

server/server/models/token.py
```diff
@@ -0,0 +1,28 @@
...
+
+    id = Column(Integer, primary_key=True, autoincrement=True)
+    token = Column(String, unique=True)
+    user_id = Column(Integer, ForeignKey('users.id'))
+    expiration_date = Column(DateTime)
+
+    user = relationship("User", foreign_keys=(user_id,))
+
+    @classmethod
+    def add_token(cls, req, token, user_id):
```
serge30: Try to do not shorten variables/parameters names, `request` is not so long. And it's the same for other methods below.

server/server/models/user.py
```diff
@@ -20,9 +23,31 @@ class User(Base):
...
+        user = req.dbsession.query(cls).filter_by(**kwargs).one_or_none()
+        return user
+
+    def check_password(self, password):
+        return pbkdf2_sha256.verify(password, self.password)
+
+    def is_active(self, req):
+        user_status = req.dbsession.query(UserStatus)\
+                        .filter_by(id=self.status_id).one()
+        if(user_status.status == "Active"):
```
serge30: Python is not JavaScript, you don't need parentheses around condition for `if` statement: [ref](https://docs.python.org/2.7/tutorial/controlflow.html#if-statements)

server/server/views/authen.py
```diff
@@ -0,0 +1,40 @@
+from passlib.totp import generate_secret
```
serge30: Pleaaaase, do not shorten modules names too. I agree that `authentication` is bit long module name, but you'd better use common shortening as `auth`.

RayMet (approved):
Астанавітесь!

server/server/views/authen.py
```diff
@@ -0,0 +1,40 @@
...
+    return response
+
+
+@view_config(route_name='log_out', renderer='json')
+def log_out(request):
+    response = {
+        'msg': "",
+        'success': False
+    }
+    if(Token.deactivate(request, request.headers['Authorization-token'])):
```
serge30: What if `Authorization-token` header is missing in request? Also we may consider using a standard [Authorization](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization) HTTP header for this.

server/config.py
```diff
@@ -1,3 +1,70 @@
...
+print run.communicate()
+
+"""
+    Seeding
+"""
+
+# seed_roles #
+
+seed_roles = config.get('seeding', 'seed_role')
+run = subprocess.Popen(seed_roles.split(), stdout=subprocess.PIPE)
```
serge30, line 25: I cannot get, why you need to call as subprocess python script, why not just to import it and run a function? You don't have to fix it in scope of this PR, but it is something we need to discuss and re-work in future.

serge30 (changes requested):

server/server/models/user.py
```diff
@@ -1,12 +1,15 @@
 """SQLAlchemy model for table users"""
+from datetime import datetime
+from passlib.hash import pbkdf2_sha256
 from sqlalchemy import Column, Integer, String, DateTime, ForeignKey
 from sqlalchemy.orm import relationship
 from . import Base
+from .user_status import UserStatus
```
serge30: According to [PEP-8](https://www.python.org/dev/peps/pep-0008/#imports):
> Imports should be grouped in the following order:
>
>Standard library imports.
>Related third party imports.
>Local application/library specific imports.
>
>You should put a blank line between each group of imports.

server/migrations/env.py
```diff
@@ -1,7 +1,9 @@
 from __future__ import with_statement
 from alembic import context
-from sqlalchemy import engine_from_config, pool
 from logging.config import fileConfig
+from sqlalchemy import engine_from_config, pool, MetaData
+from server.models import category, event_history, event_status, event_tag, event, feedback, gallery, role, subscribe, tag, user_status, user, token
```
serge30: According to [PEP-8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length):
> Limit all lines to a maximum of 79 characters.

serge30 (changes requested):

server/migrations/env.py
```diff
@@ -1,7 +1,12 @@
 from __future__ import with_statement
 from alembic import context
-from sqlalchemy import engine_from_config, pool
 from logging.config import fileConfig
+from sqlalchemy import engine_from_config, pool, MetaData
+
+from server.models import category, event_history, event_status, event_tag,\
+                        event, feedback, gallery, role,\
+                        subscribe, tag, user_status, user, token
```
serge30: PEP-8:
> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

Python ignores line ends inside brackets and parentheses, so recommendation here to wrap imports to parentheses:
```
from module import (obj1, obj2, obj3,
                    obj4, obj5)
```

serge30 (changes requested):

serge30 (approved):

PR #9, opened 2018-07-31
Feature/manage personal info

yazhyvyi:
REST API for "manage profile (user)"


server/migrations/env.py
```diff
@@ -70,13 +70,10 @@ def run_migrations_offline():
     with context.begin_transaction():
         context.run_migrations()
 
-
 def run_migrations_online():
     """Run migrations in 'online' mode.
-
```
Merteg: You did not have to delete an empty line between two parts of docstring. 
According to PEP257:
>Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description.

server/server/__init__.py
```diff
@@ -1,8 +1,8 @@
 from pyramid.config import Configurator
 
-
 def main(global_config, **settings):
     """ This function returns a Pyramid WSGI application."""
+
```
Merteg: You dont need to add a blank line after one-line docstring.
According to PEP257 - One-line Docstrings:
>There's no blank line either before or after the docstring.

server/server/__init__.py
```diff
@@ -1,8 +1,8 @@
 from pyramid.config import Configurator
```
Merteg: By PEP8 style guide you must to add two blank lines between the import statements and other code

server/server/models/__init__.py
```diff
@@ -2,11 +2,14 @@
...
+
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import scoped_session, sessionmaker, configure_mappers
+from sqlalchemy import *
 from pyramid.threadlocal import get_current_registry
 from sqlalchemy import engine_from_config
-from sqlalchemy.ext.declarative import declarative_base
-from sqlalchemy.orm import configure_mappers, sessionmaker
+from zope.sqlalchemy import ZopeTransactionExtension
+import zope.sqlalchemy
```
Merteg: t's not in PEP8, but it's common to sort all imports in alphabetical order keeping 2 groups: import module and from module import ))

server/server/models/__init__.py
```diff
@@ -32,11 +34,9 @@ def get_session_factory(engine):
 
 def get_tm_session(session_factory, transaction_manager):
     """ Get a sqlalchemy Session instance backed by a transaction.
-
```
Merteg: Please don't delete blank lines in docstrings. Not only in this file, anywhere

server/server/routes.py
```diff
@@ -3,3 +3,28 @@ def includeme(config):
     config.add_route('home', '/')
     config.add_route('log_in', '/log-in')
     config.add_route('log_out', '/log-out')
+    config.add_route(
```
Merteg: More readable and convenient will be to use Cornice for realize Rest. You don't have to fix it in scope of this pull-request, but it is something we need to re-work in future.

server/server/models/user.py
```diff
@@ -1,13 +1,39 @@
 """SQLAlchemy model for table users"""
-from datetime import datetime
+
+import json
+import decimal, datetime
```
Merteg: According to PEP8:
>Imports should usually be on separate lines

server/server/models/user.py
```diff
@@ -1,13 +1,39 @@
...
 
 from passlib.hash import pbkdf2_sha256
-from sqlalchemy import Column, Integer, String, DateTime, ForeignKey
+from sqlalchemy import Column, Integer, String, DateTime, ForeignKey, bindparam
 from sqlalchemy.orm import relationship
+from sqlalchemy.orm import Query
+from sqlalchemy.ext.declarative import DeclarativeMeta
 
 from . import Base
 from .user_status import UserStatus
```
Merteg: Expect two blank lines after imports

server/server/views/profile.py
```diff
@@ -0,0 +1,49 @@
...
+from server.models import get_dbsession
+
+from server.models import Base
+from server.models.user import User
+from server.models.user import users_to_json
+from server.models.user import datetime_conv, AlchemyEncoder
+
+@view_config(route_name='users.view', renderer='json')
+def get_users(request):
+    qry = get_dbsession().query(User).all()
```
Merteg: Better to separate logic and make all connections to db in models. You can create a method that will gets request like parameter for this.

Merteg (changes requested):
Also I don't understand AlchemyEncoder class and other conversion functions.
A DateTime field in instance of model easy converts to string by default Python method:
str_date = (str)date

requirements.txt
```diff
@@ -23,3 +23,4 @@ WebOb==1.8.2
 zope.deprecation==4.3.0
 zope.interface==4.5.0
 zope.sqlalchemy==1.0
+passlib==1.7.1
```
serge30: `passlib` is already here in line 7

server/server/models/__init__.py
```diff
@@ -2,11 +2,14 @@
 SQLAlchemy models for database
 """
 import transaction
-import zope.sqlalchemy
+
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import scoped_session, sessionmaker, configure_mappers
+from sqlalchemy import *
```
serge30: Please never use wildcard imports because it usually is a source of errors hard to debug, and even [PEP-8](https://www.python.org/dev/peps/pep-0008/#imports) asks to do not do this:
> Wildcard imports (from <module> import *) should be avoided, as they make it unclear which names are present in the namespace, confusing both readers and many automated tools. There is one defensible use case for a wildcard import, which is to republish an internal interface as part of a public API (for example, overwriting a pure Python implementation of an interface with the definitions from an optional accelerator module and exactly which definitions will be overwritten isn't known in advance).

server/server/models/user.py
```diff
@@ -69,3 +95,30 @@ def is_active(self, request):
...
+        self.create_date = create_date
+        self.location = location
+        self.first_name = first_name
+        self.last_name = last_name
+        self.status_id = status_id
+        self.role_id = role_id
+        self.avatar = avatar
+        self.banned_to_date = banned_to_date
+
+def users_to_json(obj):
```
serge30: It's a good candidate to be converted to method of `User` model.

Also according to https://stackoverflow.com/questions/1958219/convert-sqlalchemy-row-object-to-python-dict you may use something like `inspect(obj).mapper.column_attrs` or `obj.__table__.columns` (depending of SQLAlchemy version) to get column names instead of hardcoding.

server/server/models/category.py
```diff
@@ -9,7 +9,7 @@ class Category(Base):
     """SQLAlchemy model for table categories"""
 
     __tablename__ = 'categories'
-    
+
```
serge30, line 12: It's preferable to exclude changes from PR that are not to related to feature you implement. Different style fixing and code cleaning can be posted as separate PR.

server/server/views/profile.py
```diff
@@ -0,0 +1,49 @@
...
+    return json.loads(json.dumps(qry, cls=AlchemyEncoder))
+
+@view_config(route_name='users.create', renderer='json')
+def create_user(request):
+    json_str = request.json_body
+    json_encoded = json.dumps(json_str)
+    json_decoded = json.loads(json_encoded)
```
serge30: It's unclear for me what is a purpose of this actions? You takes decoded JSON from body, encodes it JSON and then decodes back. It's a bit weird.

server/server/views/profile.py
serge30: Also there already is a `dbsession`  attribute in `request` object, which automatically committed after request end.

server/server/views/profile.py
```diff
@@ -0,0 +1,49 @@
...
+    user = User(**json_decoded)
+    get_dbsession().add(user)
+    transaction.commit()
+    return {'status': 'Success!'}
+
+@view_config(route_name='users.read', renderer='json')
+def read_user(request):
+    qry = get_dbsession().query(User).filter(User.id==request.matchdict['user_id']).first()
```
serge30: There is method in SQLAlchemy to get model by id: http://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.get

serge30: Also what if id is not present in DB? You need to return correct HTTP 404 code in this case.

server/server/views/profile.py
```diff
@@ -0,0 +1,49 @@
...
+    dict_ = users_to_json(qry)
+    return json.loads(json.dumps(dict_, default=datetime_conv))
```
serge30: I really don't like this way to convert all datetime to string. It's very overcomplicated.

You may try to write a helper function which iterates over dict and convert all values which has `datetime` type to string. 
Hint: to check whether value has `datetime` type you may use: `isinstance(value, datetime.datetime)`

serge30 (changes requested):

server/server/models/user.py
```diff
@@ -69,3 +76,97 @@ def is_active(self, request):
...
+        self.first_name = first_name
+        self.last_name = last_name
+        self.status_id = status_id
+        self.role_id = role_id
+        self.avatar = avatar
+        self.banned_to_date = banned_to_date
+
+
+def from_json(cls, data):
+    """ Deserialize User object from JSON """
```
serge30: Please do not call it JSON, as JSON is a string with data in specific format. Here you are dealing with dicts.

server/server/models/user.py
```diff
@@ -69,3 +76,97 @@ def is_active(self, request):
...
+    return cls(**data)
+
+
+def model_to_json(sqlalchemy_object):
+    fields_arr = [prop.key for prop in
+                  class_mapper(sqlalchemy_object.__class__).iterate_properties
+                  if isinstance(prop, ColumnProperty)]
+    _dict = {}
+    for key in fields_arr:
+        _dict[key] = str(getattr(sqlalchemy_object, key))
```
serge30: I don't like idea to convert **all** values to str, as JSON distinguish string, float, bool and you will need to parse it in frontend side, so types might be useful. Please convert only datetime values.

server/server/models/user.py
```diff
@@ -69,3 +76,97 @@ def is_active(self, request):
...
+    return _dict
+
+
+def users_to_json(obj):
```
serge30: I believe `model_to_json` function does the same, and this one can be removed.

server/server/models/user.py
```diff
@@ -69,3 +76,97 @@ def is_active(self, request):
...
+    d = {}
+    for attr_name in to_serialize:
+        d[attr_name] = str(getattr(obj, attr_name))
+    return d
+
+
+def get_all_users():
+    """ Method to get all users from database """
+    qry = get_dbsession().query(User).all()
+    usr_in_json = list(map(lambda x: model_to_json(x), qry))
```
serge30: `map` function already returns list, anyway list comprehension is preferred:
```
a = [model_to_json(model) for model in qry]
```

server/server/models/user.py
```diff
@@ -69,3 +76,97 @@ def is_active(self, request):
...
+    return usr_in_json
+
+
+def create_usr(json_str):
```
serge30: Please do not shorten terms, `user` is not so long I blieve

server/server/models/user.py
```diff
@@ -69,3 +76,97 @@ def is_active(self, request):
...
+def read_usr(request):
+    """ Method to read user from database """
+    try:
+        qry = get_dbsession().query(User)\
+            .filter(User.id == request.matchdict['user_id']).first()
+        dict_ = users_to_json(qry)
+    except:
+        request.response.status = 404
+        return {'status': '404'}
+    return json.loads(json.dumps(dict_))
```
serge30: But why?

serge30 (changes requested):
Please take a look to Andrii's and my comments we left you before. You did a changes to the code and github collapsed them, but they are not still addressed.

server/server/models/user.py
```diff
@@ -1,43 +1,24 @@
...
 
 import json
-import decimal, datetime
+import decimal
+import datetime
+import transaction
 
 from passlib.hash import pbkdf2_sha256
 from sqlalchemy import Column, Integer, String, DateTime, ForeignKey, bindparam
 from sqlalchemy.orm import relationship
```
Merteg: You have three lines like `from sqlalchemy.orm import ...`
I think better to call all imports from one module in one line

server/server/models/user.py
```diff
@@ -109,16 +92,81 @@ def __init__(self, email, nickname, password, create_date, location, first_name,
...
+
+
 def users_to_json(obj):
     """ Serialize User object to JSON """
-
-    to_serialize = ['id', 'email', 'nickname', 'password', 'create_date', 'location', 'first_name', 'last_name', 'status_id', 'role_id', 'avatar', 'banned_to_date']
+    to_serialize = ['id', 'email', 'nickname', 'password', 'create_date',
+                    'location', 'first_name', 'last_name', 'status_id',
+                    'role_id', 'avatar', 'banned_to_date']
     d = {}
```
Merteg: If you can, name this variable dict. The single letter variable name is bad.

server/server/models/user.py
```diff
@@ -109,16 +92,81 @@ def __init__(self, email, nickname, password, create_date, location, first_name,
...
+    """ Method to get all users from database """
+    qry = get_dbsession().query(User).all()
+    usr_in_json = list(map(lambda x: model_to_json(x), qry))
+    return usr_in_json
+
+
+def create_usr(json_str):
+    """ Method to add user to database """
+    user = User(**json_str)
+    get_dbsession().add(user)
```
Merteg: Better to use request.dbsession. You can get `request` like parameter for this method. When you do this you also don't need to call `transaction.commit()`

RayMet (changes requested):
Fix problems

server/server/models/user.py
```diff
@@ -109,16 +92,81 @@ def __init__(self, email, nickname, password, create_date, location, first_name,
...
+    """ Method to update user in database """
+    json_encoded = json.dumps(json_str)
+    try:
+        get_dbsession().query(User)\
+            .filter(User.id == request.matchdict['user_id']).\
+            update(json.loads(json_encoded))
+        transaction.commit()
+    except:
+        request.response.status = 404
+        return {'status': '404'}
```
Merteg: If you want to return status 404, better to use pyramid.httpexceptions
[https://docs.pylonsproject.org/projects/pyramid/en/latest/api/httpexceptions.html](url)

Merteg (changes requested):

server/server/models/user.py
```diff
@@ -104,7 +104,11 @@ def model_to_json(sqlalchemy_object):
                   if isinstance(prop, ColumnProperty)]
     _dict = {}
     for key in fields_arr:
-        _dict[key] = str(getattr(sqlalchemy_object, key))
+        temp = getattr(sqlalchemy_object, key)
+        if isinstance(temp, datetime.datetime):
+            _dict[key] = str(getattr(sqlalchemy_object, key))
```
Merteg: You can use your `temp` variable here, it will be better than calling `getattr` method again. Also please change the `temp` name on something more informative.

server/server/models/user.py
```diff
@@ -136,37 +140,25 @@ def create_usr(json_str):
...
     """ Method to read user from database """
-    try:
-        qry = get_dbsession().query(User)\
-            .filter(User.id == request.matchdict['user_id']).first()
-        dict_ = users_to_json(qry)
-    except:
-        request.response.status = 404
-        return {'status': '404'}
-    return json.loads(json.dumps(dict_))
+    qry = get_dbsession().query(User)\
```
Merteg: Please don't use get_dbsession function. Better to use request.dbsession.

server/server/models/user.py
```diff
@@ -136,37 +140,25 @@ def create_usr(json_str):
...
+        .filter(User.id == request.matchdict['user_id']).first()
+    dict_ = users_to_json(qry)
+    return json.loads(json.dumps(dict_, cls=DTEncoder))
```
Merteg: Seems like you encode dict to the json and decode it back.

Merteg (changes requested):

server/server/models/user.py
```diff
@@ -1,19 +1,18 @@
...
-import decimal
 import datetime
-import transaction
 
 from passlib.hash import pbkdf2_sha256
 from sqlalchemy import Column, Integer, String, DateTime, ForeignKey, bindparam
 from sqlalchemy.orm import relationship
 from sqlalchemy.orm import Query
 from sqlalchemy.ext.declarative import DeclarativeMeta
 from sqlalchemy.orm import class_mapper, ColumnProperty
```
Merteg: As I wrote earlier, I think better to write `from sqlalchemy.orm import class_mapper, ColumnProperty, Query, relationship`, instead of writing from sqlalchemy.orm ... three times.

server/server/models/user.py
```diff
@@ -69,3 +72,113 @@ def is_active(self, request):
...
+
+
+def get_all_users(request):
+    """ Method to get all users from database """
+    qry = request.dbsession.query(User).all()
+    usr_in_json = list(map(lambda x: model_to_json(x), qry))
+    return usr_in_json
+
+
+def create_usr(json_str, request):
```
serge30: Please do not shorten words in function/variables, where there are reasons to do this. I believe `user` is not so much longer than `usr`.

server/server/models/user.py
```diff
@@ -69,3 +72,113 @@ def is_active(self, request):
...
+    request.dbsession.add(user)
+    return {'status': 'Success!'}
+
+
+def read_usr(request):
+    """ Method to read user from database """
+    if request.dbsession.query(User)\
+            .filter(User.id == request.matchdict['profile_id']).count():
+        qry = request.dbsession.query(User)\
+            .filter(User.id == request.matchdict['profile_id']).first()
```
serge30: I'd better use [get](http://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.get) method to take a model by id. And you don't need to check presence before fetching model. Just `get` it, and if it doesn't exist you will receive None.

server/server/views/profile.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+@resource(collection_path='/profile', path='/profile/{profile_id}',
+          renderer='json')
+class UserProfile(object):
+
+    def __init__(self, request, context=None):
+        self.request = request
+        self.context = context
+        #self.owner = request.matchdict['profile_id']
```
serge30: Do not commit commented code, just remove it.

server/server/views/profile.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def post(self):
+        request = self.request
+        json_str = request.json_body
+        return create_usr(json_str, request)
+
+    #@view(permissions='edit')
+    def put(self):
+        request = self.request
+        json_str = request.json_body
```
serge30: It's not a string, so `_str` suffix is confusing.

server/server/models/user.py
```diff
@@ -69,3 +72,113 @@ def is_active(self, request):
...
+
+
+def update_usr(json_str, request):
+    """ Method to update user in database """
+    json_encoded = json.dumps(json_str)
+    if request.dbsession.query(User)\
+            .filter(User.id == request.matchdict['profile_id']).count():
+        request.dbsession.query(User)\
+            .filter(User.id == request.matchdict['profile_id']).\
+            update(json.loads(json_encoded))
```
serge30: But why you need to encode first dict JSON and then decode it????

server/server/models/user.py
```diff
@@ -69,3 +72,113 @@ def is_active(self, request):
...
+    def default(self, obj):
+        if isinstance(obj, datetime.datetime):
+            return o.isoformat()
+        return json.JSONEncoder.default(self, obj)
+
+
+def get_all_users(request):
+    """ Method to get all users from database """
+    qry = request.dbsession.query(User).all()
+    usr_in_json = list(map(lambda x: model_to_json(x), qry))
```
serge30: `usr_in_json = [model_to_json(x) for x in qry]` is more pythonic

serge30 (changes requested):
You still have conflicts, please try to resolve them and I could finish my review.

server/server/models/user.py
```diff
@@ -92,14 +91,30 @@ def __init__(self, email, nickname, password, create_date, location,
...
+                .filter(User.id == request.matchdict['profile_id']).\
+                update(json_data)
+        else:
+            raise exc.exception_response(404)
+        return {'status': 'Success!'}
 
 
-def model_to_json(sqlalchemy_object):
-    """ Serializes sqlalchemy_object to json
+def model_to_dict(sqlalchemy_object):
```
Merteg: You can use this function for all models, so it have no reason to store this function in user.py file. (In my opinion, will be better to save this function in __init__.py at model folder)

Merteg (changes requested):
...and you must to resolve all conflicts

Merteg (approved):
I think all right

server/alembic.ini
```diff
@@ -0,0 +1,74 @@
+# A generic, single database configuration.
```
serge30: `alembic.ini` has been removed in scope of #13 Please revert this changes

server/server/views.py
```diff
@@ -0,0 +1,5 @@
+from pyramid.view import view_config, view_defaults
```
serge30: I'm not sure this file is still needed. Anyway you already have package with name `server.views`, and you are adding module with same name. There might be some conflicts on importing.

serge30 (changes requested):

yazhyvyi:
I just resolved those requested changes (maybe?)

server/server/views/profile.py
```diff
@@ -0,0 +1,35 @@
+import json
+import transaction
```
serge30: Looks like unused imports. Please review imports, as I see others unused ones,

server/server/models/user.py
```diff
@@ -1,9 +1,14 @@
 """SQLAlchemy model for table users"""
-from datetime import datetime
+
+import datetime
+import json
```
serge30: Unused import.

serge30 (changes requested):
Please check imports

server/server/models/user.py
```diff
@@ -1,18 +1,20 @@
 """SQLAlchemy model for table users"""
-from datetime import datetime
+
+import datetime
 
 from passlib.hash import pbkdf2_sha256
 from sqlalchemy import Column, Integer, String, DateTime, ForeignKey
 from sqlalchemy.orm import relationship
 
+import pyramid.httpexceptions as exc
```
serge30: I understand that `exc` is shorter than `httpexceptions`, but for readability the latter is much better:
```
from pyramid import httpexceptions
```

serge30 (changes requested):

yazhyvyi:
Sample Text

serge30 (approved):

RayMet (approved):

PR #10, opened 2018-07-31
Feature/recover pass

RayMet:


server/mailer_config
```diff
@@ -0,0 +1,22 @@
+mailer
```
Merteg: What is this? This file has no extension. Also this text exist in README

server/server/routes.py
```diff
@@ -1,5 +1,7 @@
 def includeme(config):
     config.add_static_view('static', 'static', cache_max_age=3600)
     config.add_route('home', '/')
+    config.add_route('recover_password', '/recover_password')
```
Merteg: Don't use underline in url. Better to use '-'

server/server/views/recover.py
```diff
@@ -0,0 +1,34 @@
+import transaction
+from pyramid_mailer.mailer import Mailer
+from pyramid_mailer.message import Message
+from pyramid.view import view_config
+from server.models.user_status import UserStatus
+from passlib.hash import pbkdf2_sha256
+from passlib.totp import generate_secret
```
Merteg: PEP8:
>Imports should be grouped
>You should put a blank line between each group of imports.

Also, a good practice is to sort imports in alphabetic order.

server/server/views/recover.py
```diff
@@ -0,0 +1,34 @@
...
+        token = request.dbsession.query(User).filter_by(email=jsn['email']).one_or_none()
+        mailer = request.mailer
+        message = Message(subject="hello world",
+                          sender="eventmerv@gmail.com",
+                          recipients=[jsn["email"]],
+                          body='http://localhost:6543/change_password/')
+        mailer.send(message)
+        mailer.send_immediately(message, fail_silently=False)
+
+    return "olala"
```
Merteg: This response is not very informative.

server/setup.py
```diff
@@ -12,6 +12,7 @@
     'plaster_pastedeploy',
     'pyramid',
     'pyramid_jinja2',
+    'pyramid_mailer',
```
Merteg: This module exist in requirements.txt. Don't duplicate this

server/venv/bin/activate
```diff
@@ -0,0 +1,78 @@
+# This file must be used with "source bin/activate" *from bash*
```
Merteg: Delete virtualenv folder from git repository.

Merteg (changes requested):

server/server/routes.py
serge30: Underscores in URL's path are ok. But they are not allowed in domain names.

serge30 (changes requested):
You've added whole virtual env with 1.5k files to this PR, which doesn't make sense and makes this PR unreviewable. Please leave only files needed to your feature.

README.md
```diff
@@ -25,6 +25,26 @@ Adminer:
 localhost:9000
 ```
 
+<<<<<<< HEAD
```
serge30: This is merge conflict marker which you had to remove during conflict resolving.

README.md
```diff
@@ -25,6 +25,26 @@ Adminer:
...
+    
+    mail.username  
+    
+  and
+
+    mail.password 
+
+- Set your email address in mail.username and password in mail.password.
+
+=======
```
serge30: The same as above.

server/.gitignore
```diff
@@ -19,3 +19,4 @@ Data.fs*
 .DS_Store
 coverage
 test
+venv/
```
serge30: Please leave last line blank.

server/development.ini
```diff
@@ -13,9 +13,15 @@ pyramid.debug_routematch = false
 pyramid.default_locale_name = en
 pyramid.includes =
     pyramid_debugtoolbar
-
+    
 sqlalchemy.url = postgresql://admin:1@eventme_postgres/eventme
 
+mail.host = smtp.gmail.com
+mail.username = eventmerv@gmail.com
+mail.password = Docker666Vim
```
serge30: It'a always a bad idea to store passwords in public repositories, where everyone can see them.

The idea is next:

1. Rename development.ini to development.ini.sample
2. Add development.ini to .gitignore
3. Add instruction to README to copy development.ini.sample to development.ini and fill credentails before first deployment
4. ....
5. Profit

server/server/views/recover.py
```diff
@@ -0,0 +1,35 @@
...
+from pyramid_mailer.message import Message
+from passlib.hash import pbkdf2_sha256
+from server.models.user_status import UserStatus
+from pyramid.view import view_config
+
+from ..models.user import User
+
+@view_config(route_name='recover_password', renderer='json')
+def recover_send_mail(request):
+    jsn = request.json_body
```
serge30: Please do not shorten words in variables names.

server/server/views/recover.py
```diff
@@ -0,0 +1,35 @@
...
+    user = request.dbsession.query(User).filter_by(email=jsn['email']).one_or_none()
+    if user.email is not None:
+        token = request.dbsession.query(User).filter_by(email=jsn['email']).one_or_none()
+        mailer = request.mailer
+        message = Message(subject="hello world",
+                          sender="eventmerv@gmail.com",
+                          recipients=[jsn["email"]],
+                          body='http://localhost:6543/change_password/')
+        mailer.send(message)
+        mailer.send_immediately(message, fail_silently=False)
```
serge30: Why do you send message twice?

server/server/views/recover.py
```diff
@@ -0,0 +1,35 @@
...
+from pyramid.view import view_config
+
+from ..models.user import User
+
+@view_config(route_name='recover_password', renderer='json')
+def recover_send_mail(request):
+    jsn = request.json_body
+    user = request.dbsession.query(User).filter_by(email=jsn['email']).one_or_none()
+    if user.email is not None:
+        token = request.dbsession.query(User).filter_by(email=jsn['email']).one_or_none()
```
serge30: I don't that this variable is used.

server/server/views/recover.py
```diff
@@ -0,0 +1,35 @@
...
+        mailer = request.mailer
+        message = Message(subject="hello world",
+                          sender="eventmerv@gmail.com",
+                          recipients=[jsn["email"]],
+                          body='http://localhost:6543/change_password/')
```
serge30: URLs hardcoding is not good idea as it might be changes in future. There is [Request.route_url](https://docs.pylonsproject.org/projects/pyramid/en/1.9-branch/api/request.html#pyramid.request.Request.route_url) method you need to use.

server/server/views/recover.py
```diff
@@ -0,0 +1,35 @@
...
+        mailer.send_immediately(message, fail_silently=False)
+
+    return "We send link for change password in your mail", jsn['email']
+
+@view_config(route_name='change_password', renderer='json')
+def recover_change_password(request):
+    jsn = request.json_body
+    user = request.dbsession.query(User).filter_by(email=jsn['email']).one_or_none()
+    password = request.dbsession.query(User).filter_by(password=jsn['password']).one_or_none()
+    if user.email is not None:
```
serge30: Weird method, you allows everybody to change any user password only knowing his/her e-mail. You need some security token user will receive by email and this method will check it.

server/server/views/recover.py
```diff
@@ -0,0 +1,35 @@
...
+        user = request.dbsession.query(User).filter_by(email=jsn['email']).update(dict(password=jsn['password']))
```
serge30: You cannot just set password in DB as it should be hashed before, there is a special method to set password you need to use.

server/server/views/recover.py
```diff
@@ -0,0 +1,35 @@
...
+        mailer.send(message)
+        mailer.send_immediately(message, fail_silently=False)
+
+    return "We send link for change password in your mail", jsn['email']
+
+@view_config(route_name='change_password', renderer='json')
+def recover_change_password(request):
+    jsn = request.json_body
+    user = request.dbsession.query(User).filter_by(email=jsn['email']).one_or_none()
+    password = request.dbsession.query(User).filter_by(password=jsn['password']).one_or_none()
```
serge30: Why do you need a user with password sent in request? Anyway it will return nothing as all passwords are hashed in DB.

server/server/views/recover.py
```diff
@@ -0,0 +1,35 @@
...
+from ..models.user import User
+
+@view_config(route_name='recover_password', renderer='json')
+def recover_send_mail(request):
+    jsn = request.json_body
+    user = request.dbsession.query(User).filter_by(email=jsn['email']).one_or_none()
+    if user.email is not None:
+        token = request.dbsession.query(User).filter_by(email=jsn['email']).one_or_none()
+        mailer = request.mailer
+        message = Message(subject="hello world",
```
serge30: I don't think it's a correct subject.

server/server/views/recover.py
```diff
@@ -0,0 +1,35 @@
...
+                          sender="eventmerv@gmail.com",
```
serge30: It's a good candidate to go to settings. You may change e-mail in future and will have to update sources, which is not good. To reach settings you can use `request.registry.settings`. ([additional info](https://docs.pylonsproject.org/projects/pyramid/en/1.9-branch/narr/environment.html#adding-a-custom-setting))

serge30 (changes requested):

server/development.ini
RayMet: Can I just make empty fields?

serge30: No, it's not enough. As every developer after updating those fields will have this file marked as changed by git and can accidentally commit it.

server/server/views/recover.py
```diff
@@ -0,0 +1,35 @@
...
+def recover_send_mail(request):
+    json = request.json_body
+    user = request.dbsession.query(User).filter_by(email=json['email']).one_or_none()
+    if user.email is not None:
+        token = request.dbsession.query(User).filter_by(email=json['email']).one_or_none()
+        mailer = request.mailer
+        message = Message(subject="hello world",
+                          sender="eventmerv@gmail.com",
+                          recipients=[json["email"]],
+                          body='http://localhost:6543/change_password/')
```
Merteg: Better to get URL by pyramid method instead of typing it like this

server/server/views/recover.py
```diff
@@ -0,0 +1,35 @@
...
+from passlib.hash import pbkdf2_sha256
+from server.models.user_status import UserStatus
+from pyramid.view import view_config
+
+from ..models.user import User
+
+@view_config(route_name='recover_password', renderer='json')
+def recover_send_mail(request):
+    json = request.json_body
+    user = request.dbsession.query(User).filter_by(email=json['email']).one_or_none()
```
Merteg: Please don't make queries to db from view. Better to separate logic. You can create method in model that will be get request object like parameter.

server/server/views/recover.py
```diff
@@ -0,0 +1,35 @@
...
+    if user.email is not None:
+        token = request.dbsession.query(User).filter_by(email=json['email']).one_or_none()
```
Merteg: You have this object in variable user in line 14, and I don't understand, why it named `token` here

server/server/views/recover.py
```diff
@@ -0,0 +1,35 @@
...
+        mailer = request.mailer
+        message = Message(subject="hello world",
+                          sender="eventmerv@gmail.com",
+                          recipients=[json["email"]],
+                          body='http://localhost:6543/change_password/')
+        mailer.send(message)
+        mailer.send_immediately(message, fail_silently=False)
+
+    return "We send link for change password in your mail", json['email']
```
Merteg: You need add a space after word `mail`, because the response will be something like this:
 "We send link for change password in your mailexample@gmail.com"

Merteg (changes requested):
You have not fixed all the problems, that were described in previous comments.

server/server/views/recover.py
```diff
@@ -8,28 +10,42 @@
...
                           sender="eventmerv@gmail.com",
                           recipients=[json["email"]],
-                          body='http://localhost:6543/change_password/')
+                          body='temporary password:'+ temporary_password +' http://localhost:6543/change_password/')
         mailer.send(message)
         mailer.send_immediately(message, fail_silently=False)
 
     return "We send link for change password in your mail", json['email']
 
+@resource(path='change_password', renderer='json')
```
Merteg: You don't need to define resource for one route with one request method. Better to use service for this. Also if you use Cornice, why did you create view by `@view_config`?

Merteg (changes requested):
I think you don't finish this task and you don't fix all problems. Cornice docs can help you a lot.

RayMet:
Change functions recover_password and change_password.

server/server/__init__.py
```diff
@@ -6,6 +6,7 @@ def main(global_config, **settings):
 
     config = Configurator(settings=settings)
     config.include('pyramid_jinja2')
+    config.include('pyramid_mailer')
```
serge30: `pyramid_mailer` is already included in line 11

server/server/views/recover.py
```diff
@@ -0,0 +1,59 @@
...
+def recover_change_password(request):
+    """Change on new password view
+
+    This function get email and new password from json request,
+    function get url_token, check if it in db.
+    Check if this email and token are in db,
+    after change password and delete value from url_token.
+    """
+    json = request.json_body
+    user_change_password = request.matchdict['change_password_hash']
```
serge30: Why does variable at left have so strange name? It doesn't reflect what it contains.

server/server/views/recover.py
```diff
@@ -0,0 +1,59 @@
...
+    user = request.dbsession.query(User).filter_by(email=json['email']).one_or_none()
+    if user.email is not None and user.url_token == user_change_password:
```
serge30: Why do you check `email` field here, it cannot be empty, as you take user model filtering by email from JSON request. But `user` variable could, when the re no user with such email, and you will have error here.

server/server/views/recover.py
```diff
@@ -0,0 +1,59 @@
...
+    json = request.json_body
+    user = request.dbsession.query(User).filter_by(email=json['email']).one_or_none()
+    if (user.email is not None) and user.is_active(request):
+        url_token_confirmation = generate_secret()
+        user.url_token = url_token_confirmation
+        mailer = request.mailer
+        message = Message(subject="Recover password",
+                          sender="eventmerv@gmail.com",
+                          recipients=[json["email"]],
+                          body='Follow the link below' + '\n' + request.route_url('change_password',
```
serge30:
```
'Follow the link below' + '\n' + ....
```
why do you need such concatenation? `'Follow the link below\n' +` is absolutely enough here.

server/server/views/recover.py
```diff
@@ -0,0 +1,59 @@
+import transaction
+from cornice import Service
+from passlib.totp import generate_secret
+from pyramid_mailer.mailer import Mailer
+from pyramid_mailer.message import Message
+from passlib.hash import pbkdf2_sha256
+from passlib.totp import generate_secret
```
serge30: It's already imported in line 3. Also I see here many unused imports, review them and remove extra ones.

serge30 (changes requested):

server/server/views/recover.py
```diff
@@ -0,0 +1,59 @@
...
+    Check if this email and token are in db,
+    after change password and delete value from url_token.
+    """
+    json = request.json_body
+    user_change_password = request.matchdict['change_password_hash']
+    user = request.dbsession.query(User).filter_by(email=json['email']).one_or_none()
+    if user.email is not None and user.url_token == user_change_password:
+        user.password=pbkdf2_sha256.hash(json['password'])
+        user.url_token = None
+    return {'status': 'OK'}
```
Merteg: If you want to return status, better to return `{'success': true/false}`
The word "success" is not a rule, but better to use it because I did this earlier and you too.

server/server/views/recover.py
RayMet: So that check email not needed here?

RayMet: And user.url_token can also have error, if user variable don't see User with such email that have our JSON?

server/server/views/recover.py
```diff
@@ -0,0 +1,57 @@
...
+    Check if this email and token are in db,
+    after change password and delete value from url_token.
+    """
+    json = request.json_body
+    change_password_url_token = request.matchdict['change_password_hash']
+    user = request.dbsession.query(User).filter_by(email=json['email']).one_or_none()
+    if request.dbsession.query(User).filter_by(email=json['email']) and request.dbsession.query(User).filter_by(url_token=change_password_url_token):
+        user.password=pbkdf2_sha256.hash(json['password'])
+        user.url_token = None
+    return {'success': True/False}
```
serge30: What does it mean??? How did you test this code??? 
```
In [5]: True/False
---------------------------------------------------------------------------
ZeroDivisionError                         Traceback (most recent call last)
<ipython-input-5-45a119b70a2a> in <module>()
----> 1 True/False

ZeroDivisionError: integer division or modulo by zero
```

server/server/views/recover.py
```diff
@@ -0,0 +1,57 @@
...
+
+    This function get email and new password from json request,
+    function get url_token, check if it in db.
+    Check if this email and token are in db,
+    after change password and delete value from url_token.
+    """
+    json = request.json_body
+    change_password_url_token = request.matchdict['change_password_hash']
+    user = request.dbsession.query(User).filter_by(email=json['email']).one_or_none()
+    if request.dbsession.query(User).filter_by(email=json['email']) and request.dbsession.query(User).filter_by(url_token=change_password_url_token):
```
serge30: What are you testing here? is in DB any user with email specified and is there any user with token? but not is there a user with both those field equal to specified.

server/server/views/recover.py
```diff
@@ -0,0 +1,57 @@
+from cornice import Service
+from pyramid_mailer.mailer import Mailer
+from pyramid_mailer.message import Message
+from passlib.hash import pbkdf2_sha256
+from passlib.totp import generate_secret
+from pyramid.view import view_config
```
serge30: import are not sorted in alphabetical order

server/server/views/recover.py
```diff
@@ -0,0 +1,57 @@
...
+    user = request.dbsession.query(User).filter_by(email=json['email']).one_or_none()
+    if (user.email is not None) and user.is_active(request):
+        url_token_confirmation = generate_secret()
+        user.url_token = url_token_confirmation
+        mailer = request.mailer
+        message = Message(subject="Recover password",
+                          sender="eventmerv@gmail.com",
+                          recipients=[json["email"]],
+                          body='Follow the link below\n' + request.route_url('change_password',
+                                                                                  change_password_hash=url_token_confirmation))
```
serge30: wrong identation

server/server/views/recover.py
serge30: yep, so you need to check `user` before checking any its fields.

server/server/views/recover.py
```diff
@@ -0,0 +1,59 @@
...
+    after change password and delete value from url_token.
+    """
+    json = request.json_body
+    change_password_url_token = request.matchdict['change_password_hash']
+    user = request.dbsession.query(User).filter_by(email=json['email']).one_or_none()
+    if (user is not None
+            and user.url_token == change_password_url_token):
+        user.password=pbkdf2_sha256.hash(json['password'])
+        user.url_token = None
+    return {'success': True}
```
serge30: You always return success, but user with specified e-mail may miss or sent toke is incorrect, so in this case `{'success': False}` is useful.

server/server/views/recover.py
```diff
@@ -0,0 +1,59 @@
...
+        message = Message(
+            subject="Recover password",
+            sender="eventmerv@gmail.com",
+            recipients=[json["email"]],
+            body='Follow the link below\n' + request.route_url('change_password',
+                                                               change_password_hash=url_token_confirmation))
+        mailer.send_immediately(message, fail_silently=False)
+
+    return {
+        'msg': "We send link for change password in your mail "+ json['email'],
```
serge30: What if user was not found in DB and link was not sent? In this case we should return `success: False` with clarification.

serge30 (changes requested):

RayMet:
Add msg if user not founded

serge30 (approved):

Merteg (approved):

PR #11, opened 2018-08-01
Feature/authorization

Merteg:
Now, all you need for protection your resources, it's create a resource class and implement __acl__ method. This method must generate a sequence of ACE tuples.
> An access control entry (ACE) is a three-tuple that describes three things: an action (one of either Allow or Deny), a principal (a string describing a user or group), and a permission. For example the ACE, (Allow, 'bob', 'read').

### DIFF
--- a/server/server/models/token.py
+++ b/server/server/models/token.py
@@ -22,6 +22,16 @@ class Token(Base):
     user = relationship("User", foreign_keys=(user_id,))
 
     @classmethod
+    def get_user_id(cls, request):
+        if 'Authorization' in request.headers:
+            token = request.headers['Authorization'].split(' ')[1]
+            token_obj = request.dbsession.query(cls)\
+                        .filter_by(token=token).one_or_none()
+            if token_obj is None:
+                return None
+            return token_obj.user_id
+
+    @classmethod
     def add_token(cls, request, token, user_id):
         """Add authorization token to the db"""
         token_obj = cls(token=token, user_id=user_id,
--- a/server/server/models/token.py
+++ b/server/server/models/token.py
@@ -22,14 +22,14 @@ class Token(Base):
     user = relationship("User", foreign_keys=(user_id,))
 
     @classmethod
-    def get_user_id(cls, request):
+    def get_token_obj(cls, request):
         if 'Authorization' in request.headers:
             token = request.headers['Authorization'].split(' ')[1]
             token_obj = request.dbsession.query(cls)\
                         .filter_by(token=token).one_or_none()
             if token_obj is None:
                 return None
-            return token_obj.user_id
+            return token_obj
 
     @classmethod
     def add_token(cls, request, token, user_id):
@@ -44,3 +44,6 @@ class Token(Base):
         """Delete authorization token from db"""
         request.dbsession.query(cls).filter_by(token=token).delete()
         return True
+
+    def update(self):
+        self.expiration_date = datetime.now() + timedelta(days=TOKEN_LIFETIME)
--- a/server/server/models/token.py
+++ b/server/server/models/token.py
@@ -22,14 +22,12 @@ class Token(Base):
     user = relationship("User", foreign_keys=(user_id,))
 
     @classmethod
-    def get_token_obj(cls, request):
-        if 'Authorization' in request.headers:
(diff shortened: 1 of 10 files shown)


server/server/models/token.py
```diff
@@ -21,6 +21,16 @@ class Token(Base):
 
     user = relationship("User", foreign_keys=(user_id,))
 
+    @classmethod
+    def get_user_id(cls, request):
+        if 'Authorization' in request.headers:
+            token = request.headers['Authorization'].split(' ')[1]
```
serge30: It would be better to send already extracted token to this method instead of `request`

server/server/models/user.py
```diff
@@ -69,3 +70,9 @@ def is_active(self, request):
                                 .filter_by(status="Active").one().id
                 return True
         return False
+
+    def get_role(self, request):
+        """Return string with user role"""
+        role_obj = request.dbsession.query(Role).get(self.role_id)
```
serge30: You need to use SQLAlchemy's capabilities to handle relations. There is `roles = relationship("Role", foreign_keys=(role_id,))` (but I have no idea why it's in plural form) and you should use it.

serge30 (changes requested):

server/server/security.py
```diff
@@ -0,0 +1,67 @@
...
+from .models.user import User
+from .models.token import Token
+
+
+@implementer(IAuthenticationPolicy)
+class MyAuthenticationPolicy(CallbackAuthenticationPolicy):
+
+    def unauthenticated_userid(self, request):
+        """Returns user's id by token"""
+        userid = Token.get_user_id(request)
```
serge30: Do not forget to update token's expiration time on login.

serge30 (changes requested):

server/server/models/token.py
Merteg: I need a `request` in this method, because I need a connection to db for getting `user_id` by token.

RayMet (approved):
Nice

server/server/models/token.py
serge30: Yep, but the main idea that `Token` model should only know about how to store tokens in DB, but know nothing about way how to extract token from `request`. So it's a good to send `request` for DB session, and string with already extracted token.

Merteg: Okey

server/server/models/token.py
```diff
@@ -22,14 +22,12 @@ class Token(Base):
...
-                        .filter_by(token=token).one_or_none()
-            if token_obj is None:
-                return None
-            return token_obj
+    def get_token_obj(cls, request, token):
+        token_obj = request.dbsession.query(cls)\
+                    .filter_by(token=token).one_or_none()
+        if token_obj is None:
+            return None
+        return token_obj
```
serge30: I believe:
```
return token_obj
```
does absolutely the same what does:
```
if token_obj is None:
    return None
return token_obj
```

server/server/security.py
```diff
@@ -0,0 +1,70 @@
...
+        """Generate random authorization token, save it into db and return"""
+        key = generate_secret()
+        Token.add_token(request, key, userid)
+        return key
+
+    def forget(self, request):
+        """Delete token from db if it exist"""
+        if 'Authorization' in request.headers:
+            if Token.deactivate(request,
+                                request.headers['Authorization'].split(' ')[1]
```
serge30: You use the same snippet at least twice, so probably you may extract to the function, which returns token extarcted from request.

serge30 (changes requested):

server/server/models/token.py
```diff
@@ -29,8 +35,11 @@ def add_token(cls, request, token, user_id):
...
-    @classmethod
-    def deactivate(cls, request, token):
+    def deactivate(self, request):
         """Delete authorization token from db"""
-        request.dbsession.query(cls).filter_by(token=token).delete()
+        request.dbsession.query(self.__class__)\
+            .filter_by(token=self.token).delete()
         return True
+
+    def update(self):
```
serge30: `update` is not good for this method as it updates only one field, so it'd better to call it `update_expiration_date`.

serge30 (changes requested):

serge30 (approved):

PR #12, opened 2018-08-02
Feature/react init

Merteg:


server/server/__init__.py
```diff
@@ -6,6 +6,7 @@ def main(global_config, **settings):
     config = Configurator(settings=settings)
     config.include('pyramid_jinja2')
     config.include('.models')
+    config.include('.security')
```
serge30: Why is it needed? As I understand `security.py` is empty.

serge30 (changes requested):

serge30 (changes requested):
There is one strange action:
```
client/.geetkeep → server/server/security.py
```
Also you have empty `client/.geetkeep` in this PR. I'm not sure all of them are needed.

serge30 (approved):

PR #14, opened 2018-08-09
Feature/authentication

Merteg:


Merteg:
Wait for your review

client/src/components/Home.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { Row, Col } from 'react-materialize'
+
+
+export function Home() {
```
Merteg: Better to use default export in end of the file

client/src/components/LogOut.jsx
```diff
@@ -0,0 +1,29 @@
...
+      }
+      else {
+        this.setState({ msg: data.msg })
+      }
+      window.Materialize.toast(this.state.msg, 3000)
+    })
+  }
+
+  render() {
+    return <NavItem onClick={this.log_out} href="log-out">Log Out</NavItem>
```
Merteg: I'm not sure that I really need a `href` attribute.

client/public/index.css
```diff
@@ -0,0 +1,13 @@
+.nav-wrapper {
+    padding: 0 15px;
+}
+.input-field label{
```
serge30: As I understand you overrides this classes for input labels which affects all forms on the application, but those changes only need for login form on the header. In this case you need restrict this only for class of header.

client/src/components/Nav.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+
+export class Nav extends Component {
+  constructor(props) {
+    super(props);
+    this.update = this.update.bind(this);
+  }
+
+  isLogged() {
+    return typeof (sessionStorage['Authorization-token']) !== 'undefined' ? true : false
```
serge30: `typeof (sessionStorage['Authorization-token']) !== 'undefined'` already returns true or false, you don't need here ternary    operator

serge30 (changes requested):

client/src/components/Nav.jsx
Merteg: Also I think move this function into scripts.js file, is it ok?

serge30: no objections from my side

Merteg:
I made a change

client/src/scripts.js
```diff
@@ -0,0 +1,26 @@
+import { server_url } from './config.json'
```
serge30: I'm not sure `scripts.js` is a good name for this module, something like `utils.js` sounds better.

serge30 (changes requested):

Merteg:
Add new commit

serge30 (approved):
Looks good to me. But let's give a chance to team to review and provide feedback.

RayMet (approved):

ykobuk (approved):

yazhyvyi (approved):

PR #15, opened 2018-08-12
Implement method to add event into the system

Merteg:
Impement method for adding new event on back-end.
Add UI and method for adding event on client.
Add validation, refactor and test.


server/server/views/add_event.py
```diff
@@ -0,0 +1,83 @@
...
+    """Factory for adding ACL to the Service"""
+
+    def __init__(self, request, context=None):
+        self.request = request
+
+    def __acl__(self):
+        return [(Allow, Authenticated, 'add')]
+
+
+add_event = Service(name='add_event', path='/add-event',
```
serge30: Why don't you use cornice's [Resource](http://cornice.readthedocs.io/en/latest/resources.html) here? Event is more like resource than service.

client/src/utils.js
```diff
@@ -1,26 +1,41 @@
...
-    return /.+@{1}.+/.test(email);
-}
+export const emailValidation = email => /.+@{1}.+/.test(email);
 
-export function isLogged() {
-  return typeof (sessionStorage['Authorization-token']) !== 'undefined'
-};
+export const isLogged = () => (typeof(sessionStorage['Authorization-token']) !== 'undefined');
+
+export function dateFromString(date_str) {
```
serge30: [moment.js](https://momentjs.com/) is very good JS library to work with datetimes, and it can convert dates from [strings](https://momentjs.com/docs/#/parsing/string/). You may try.

Merteg:
Add moment.js, change service to resource.

server/server/views/create_event.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+@add_event.get()
+def add_event_get(request):
+    response = {
+        'success': False,
+        'categories': None
+    }
+    categories = Category.get_all(request)
+    response['categories'] = [model_to_dict(obj) for obj in categories]
```
serge30: If you need a list of categories, create a cornice's Resource for Category and use it.

Merteg:
Create a cornice's Resource for Category

ykobuk (approved):

RayMet (approved):

PR #18, opened 2018-08-13
Feature/profile ui

yazhyvyi:


client/src/components/Home.jsx
```diff
@@ -1,13 +1,25 @@
...
-function Home() {
-  return (
-    <Row>
-      <Col s={4} offset="s2"><h1>This is Home</h1></Col>
-    </Row>
-  );
-}
+export default class Home extends Component {
+  componentDidMount() {
+    fetch(server_url)
```
Merteg: This useless.

client/src/components/Nav/LogIn.jsx
```diff
@@ -25,6 +25,7 @@ class LogIn extends Component {
     }
     request('/log-in', 'POST', JSON.stringify(log_in_data))
     .then(data=>{
+      console.log(data)
```
Merteg: Please delete all console.log

client/src/components/Nav/NavProfile.jsx
```diff
@@ -11,12 +11,20 @@ class NavProfile extends Component {
     else return '/person.jpg'
   }
 
+  constuctor() {
+    this.routeChange = this.routeChange.bind(this);
+   }
+
+   routeChange(){
```
Merteg: I know better way to do this, Sorry but seems really bad

client/src/components/Profile.jsx
```diff
@@ -0,0 +1,150 @@
...
+        };
+
+        this.UpdateClick = this.UpdateClick.bind(this);
+        this.onChangeHandler = this.onChangeHandler.bind(this);
+        this.handleClick_Main = this.handleClick_Main.bind(this);
+        this.handleClick_Edit = this.handleClick_Edit.bind(this);
+    }
+
+    componentDidMount() {
+        fetch(server_url + '/profile/' + window.location.href.split("/").pop())
```
Merteg: Why you don't use request function from utils.js?

client/src/components/Profile.jsx
```diff
@@ -0,0 +1,150 @@
...
+        .then(response => response.json())
+        .then(json => {
+            console.log(json);
+            this.setState({ user: json });
+        })
+    }
+
+    componentWillUnmount() {
+        clearInterval(this.timer);
```
Merteg: What is this?

client/src/components/Profile.jsx
```diff
@@ -0,0 +1,150 @@
...
+          <h1>Profile</h1>
+        </div>
+        <ColoredLine color="teal" />
+        {page}
+      </div>
+    );
+  }
+}
+
+const ColoredLine = ({ color }) => (
```
Merteg: I don't think you need to define a new component. Also NEVER write the css properties here. Use .css file for this

client/src/components/Profile.jsx
```diff
@@ -0,0 +1,150 @@
...
+    handleClick_Main() {
+        window.location.reload();
+    }
+    handleClick_Edit() {
+        this.setState({Edit: true});
+    }
+
+  render() {
+      let avatar_image;
+      if (this.state.user.avatar == null) {
```
Merteg: Please remove all statements from render, better to create new method (smth like avatarImage(){})

client/src/components/Profile.jsx
```diff
@@ -0,0 +1,150 @@
...
+
+    }
+
+    onChangeHandler = (e) => {
+        let { id } = e.currentTarget
+        this.setState({ [id]: e.currentTarget.value })
+    }
+
+    handleClick_Main() {
+        window.location.reload();
```
Merteg: when you change the `state`, the component instantly re-render. You don't need to reload whole page.

client/src/components/Profile.jsx
```diff
@@ -0,0 +1,150 @@
...
+import { Row, Col, Input, Button } from 'react-materialize';
+import { server_url } from '../config.json';
+
+
+export default class Profile extends Component {
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            Edit: false,
```
Merteg: why uppercase?

client/src/components/Profile.jsx
```diff
@@ -0,0 +1,150 @@
...
+            msg: '',
+            user: [],
+        };
+
+        this.UpdateClick = this.UpdateClick.bind(this);
```
Merteg: You don't pass this methods to another components so you don't need to bind this (and call constructor too)

client/src/index.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
-import App from './App';
+import App from './App.jsx';
```
Merteg: You don't need to write extension of the component

requirements.txt
```diff
@@ -4,6 +4,7 @@ freeze==1.0.10
 hupper==1.3
 Mako==1.0.7
 MarkupSafe==1.0
+npm==0.1.1
```
Merteg: WHAT?

client/src/components/Home.jsx
```diff
@@ -1,13 +1,25 @@
...
+      .then(data => {
+        console.log('Data: ', data)
+      })
+  }
 
-export default Home;
+  render() {
+    return (
+      <Row>
+        <Col s={1}></Col>
```
RayMet: Why first "Col" is empty?

server/development.ini.sample
```diff
@@ -1,68 +0,0 @@
-###
```
Merteg: We all need this file, please don't delete it

RayMet (changes requested):

server/server/models/user.py
```diff
@@ -81,9 +81,10 @@ def get_role(self):
         return self.roles.role
 
     @staticmethod
-    def update_user(json_data, request):
+    def update_user(request, json_data):
         """ Method to update user in database """
-        if request.dbsession.query(User).get(request.matchdict['profile_id']):
```
Merteg: this line is not longer than 79 symbols

server/server/views/home.py
```diff
@@ -0,0 +1,20 @@
+from cornice.resource import resource, view
+from pyramid.security import Allow, Everyone
+
+
+@resource(path='/', renderer='json', cors_origins=('*',))
```
Merteg: Please don't allow cors to all, you need to allow it just for client

server/server/views/home.py
```diff
@@ -0,0 +1,20 @@
...
+class Home(object):
+
+    def __init__(self, request, context=None):
+        self.request = request
+        self.context = context
+
+    def __acl__(self):
+        return [(Allow, Everyone, 'everything')]
```
Merteg: 'everything' sounds really bad, pyramid have ALL_PERMISSION const for this

server/server/views/profile.py
```diff
@@ -1,3 +1,12 @@
+import json
+import transaction
+
+from cornice.resource import resource, view
+from cornice import Service
```
Merteg: Remove all unused imports

server/server/views/profile.py
```diff
@@ -6,23 +15,29 @@
...
 
-    def __init__(self, request, context=None):
+    def __init__(self, request, context=None,):
         self.request = request
         self.context = context
 
     def __acl__(self):
         return [(Allow, Everyone, ALL_PERMISSIONS),
-                (Allow, self.owner, 'edit')]
+                (Allow, Authenticated, 'edit')]
```
Merteg: It means that the every authenticated user can edit every profile page. It does not have to work like that

Merteg (changes requested):

client/src/components/Home.jsx
yazhyvyi: ok, i'll correct that

client/src/components/Profile.jsx
yazhyvyi: I wrote this piece of code before utils.js was implemented

client/src/components/Profile.jsx
yazhyvyi: Well, forgot to remove that

yazhyvyi:
Update

ykobuk (approved):

client/src/components/Nav/NavProfile.jsx
yazhyvyi: Care to elaborate?

server/server/views/profile.py
```diff
@@ -1,28 +1,38 @@
...
-        return model_to_dict(user)
+        user_dict = model_to_dict(user)
+        user_dict['password'] = None
+        return user_dict
 
+    @view(permission='edit')
     def put(self):
         request = self.request
         data = request.json_body
+        if 'password' in data is not "":
```
serge30: First, I was going to write that this statement returns incorrect result, but after experiments in console it works. What is weird.

Anyway correct statement to check key presence in the dict is `if 'password' in data`.

serge30: or if you need to check whether key is present in dict and value is not empty, you may try: `if dict.get('password'):`

yazhyvyi: Nope, returns error: 
TypeError: descriptor 'get' requires a 'dict' object but received a 'str'

Edit: Nevermind, it works, mistake on my side

serge30: I provide a wrong example, you need to use your variable:
```
if data.get('password'):
```

client/src/components/Nav/NavProfile.jsx
Merteg: `<li><Link to={"/profile/" + sessionStorage['User-id']}>` seems better and also takes less code and resources

client/src/components/Profile.jsx
Merteg: But now it's here, it takes less space and seems better. Also if we will need to upgrade it, better to do this in one place.

client/src/components/Profile.jsx
```diff
@@ -0,0 +1,144 @@
...
+        this.handleClick_Main = this.handleClick_Main.bind(this);
+        this.handleClick_Edit = this.handleClick_Edit.bind(this);
+    }
+
+    componentDidMount() {
+        fetch(server_url + '/profile/' + window.location.href.split("/").pop())
+        .then(response => response.json())
+        .then(json => {
+            this.setState({ user: json });
+            console.log(json)
```
Merteg: Remove console.log...

client/src/components/Profile.jsx
Merteg: Also I don't understand `window.location.href.split("/").pop()`, you define this route like ` <Route path="/profile/:profile_id" component={Profile}/>` so the path parameter stored in this.props.match.params.profile_id

client/src/components/Routes.jsx
```diff
@@ -1,11 +1,13 @@
 import React from 'react';
 import { Route } from 'react-router-dom';
-import Home from './Home';
+import Home from './Home.jsx';
```
Merteg: Why do you type the extension of files???

server/server/views/home.py
```diff
@@ -0,0 +1,20 @@
...
+class Home(object):
+
+    def __init__(self, request, context=None):
+        self.request = request
+        self.context = context
+
+    def __acl__(self):
+        return [(Allow, Everyone, ALL_PERMISSIONS)]
+
+    @view(permission=ALL_PERMISSIONS)
```
Merteg: Useless. We don't need any requests on the home page and this code also do nothing useful.

server/server/views/profile.py
```diff
@@ -1,28 +1,38 @@
-from cornice.resource import resource
+import json
```
Merteg: Do you use this import?

Merteg (changes requested):

client/src/components/Profile.jsx
YaZhyvyi: About NavProfile's routeChange:

I just tested your solution while being In the profile itself and it returns this error:
Warning: Hash history cannot PUSH the same path; a new entry will not be added to the history stack

client/src/components/Nav/NavProfile.jsx
```diff
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import { NavItem } from 'react-materialize'
 import {isLogged} from '../../utils'
+import {Link} from 'react-router-dom'
```
Merteg: In accordance to style code [Add spaces inside curly braces](https://github.com/airbnb/javascript/#whitespace--in-braces)

server/server/views/home.py
```diff
@@ -1,5 +1,4 @@
 from cornice.resource import resource, view
```
Merteg: We don't have this file in views and we don't need it right now. Please remove this.

server/server/views/profile.py
```diff
@@ -1,28 +1,36 @@
...
         self.request = request
         self.context = context
+        self.owner_id = int(request.matchdict['profile_id'])
 
     def __acl__(self):
         return [(Allow, Everyone, ALL_PERMISSIONS),
-                (Allow, self.owner, 'edit')]
+                (Allow, self.owner_id, 'edit')]
 
+    @view(permission=ALL_PERMISSIONS)
```
Merteg: It not a problem but if you allow it for everyone, you don't need to add permission at all.
I had to say this in the previous comment, sorry.

server/server/views/home.py
```diff
@@ -0,0 +1,15 @@
+from cornice.resource import resource, view
```
Merteg: Maybe I typed this comment earlier, nevermind...
We don't need this file at all, we don't make any requests to this right now

client/src/components/Profile.jsx
```diff
@@ -0,0 +1,143 @@
...
+        );
+      }
+      else {
+          return (
+            <div>
+              <div>
+                <h1>Profile</h1>
+              </div>
+                <Row>
+                    <Input s={12} id="email" label="Email" placeholder={this.state.user.email} onChange={this.onChangeHandler} />
```
Merteg: The input at Navbar have id="email" too. Id must be unique, you can name it smth like email_profile...

client/src/components/Profile.jsx
```diff
@@ -0,0 +1,143 @@
...
+          return (
+            <div>
+              <div>
+                <h1>Profile</h1>
+              </div>
+                <div id="main">
+                <Row>
+                    <Col s={6}>
+                        <Row>
+                            <this.Avatar_image />
```
Merteg: Seems strange, you can call it like a simple method {this.avatar_image}
Also better to name method smth like `renderAvatarImage`.
Better to use camelCase in JS [style code](https://github.com/airbnb/javascript/#naming--camelCase)

Merteg (changes requested):
It seems good. I type a comments about minor but I think important changes.

yazhyvyi:
Update

server/server/views/home.py
```diff
@@ -1,7 +1,3 @@
-from cornice.resource import resource, view
-
-
-@resource(path='/', renderer='json', cors_origins=('http://localhost:3000',))
 class Home(object):
```
Merteg: Remove. This. File.
Just DELETE.

client/src/components/Profile.jsx
```diff
@@ -12,7 +12,7 @@ export default class Profile extends Component {
             user: []
         };
 
-        this.Avatar_image = this.Avatar_image.bind(this);
+        this.renderAvatarImage = this.renderAvatarImage.bind(this);
         this.UpdateClick = this.UpdateClick.bind(this);
         this.onChangeHandler = this.onChangeHandler.bind(this);
         this.handleClick_Main = this.handleClick_Main.bind(this);
```
Merteg: This don't have any sense here

Merteg (changes requested):

yazhyvyi:
Update

client/src/components/Profile.jsx
```diff
@@ -83,7 +83,7 @@ export default class Profile extends Component {
                 <Row>
                     <Col s={6}>
                         <Row>
-                            <renderAvatarImage />
+                            <this.renderAvatarImage />
```
Merteg: {this.renderAvatarImage()}

Merteg (changes requested):
Also you don't change fetch to request

yazhyvyi:
Update

Merteg:
Fetch - request??? If you think that write `fetch` is better tell me why?

yazhyvyi:
Update

Merteg (approved):

PR #19, opened 2018-08-17
Fix/sprint2 andrew

Merteg:


server/server/views/event.py
```diff
@@ -48,18 +50,23 @@ def collection_post(self):
...
+            new_event.category_id = Category\
+                .get_by_name(request, request.validated['category'].lower()).id
+
             Event.add_event(request, obj=new_event)
             new_event_id = Event.get_event_obj(request,
                                                name=request.validated['name']
                                                ).id
+
+            if 'tags' in request.validated:
+                for tag in request.validated['tags']:
```
serge30: `dict` has very useful method [get](https://docs.python.org/2/library/stdtypes.html#dict.get). So you may substitute 2 lines above by one:
```
for tag in request.validated.get('tags', []):
```

Merteg:
Update

yazhyvyi (commented):
"Add Event": you need to make "main_image" a null value by default, it gives error on button click -- "Add Event"

Merteg:
Sorry for the delay, update

ykobuk (approved):

yazhyvyi (approved):
Did some testing, no issues here. Any additions or changes should be applied to the next PR

PR #20, opened 2018-08-17
Feature/profile ui fix

yazhyvyi:


client/src/components/ProfileEdit.jsx
```diff
@@ -0,0 +1,98 @@
...
+        }
+      })
+		})
+	}
+
+	UpdateClick = (eve) => {
+		eve.preventDefault()
+
+    let nick_error = ""
+    let repeated_error = ""
```
serge30: Please do not use tabs, and replace them by spaces. It makes hard to indent lines properly.

server/server/validation_schema.py
```diff
@@ -26,3 +26,22 @@ class EventSchema(colander.MappingSchema):
     main_image = colander.SchemaNode(colander.String(), missing=colander.drop)
     category = colander.SchemaNode(colander.String(),
                                    validator=colander.Length(max=255))
+
+
+class ProfileSchema(colander.MappingSchema):
+    """Schema for profile edit validation"""
+    first_name = colander.SchemaNode(colander.String(),
+                                 validator=colander.Length(max=255),
+                                 missing=colander.drop)
```
serge30: wrong identation

client/src/components/Profile.jsx
```diff
@@ -1,127 +1,81 @@
...
-
-    handleClick_Edit = () => {
-        this.setState({ edit: true });
-    };
+  componentDidMount() {
+      this.setState({ page_id: this.props.match.params.profile_id });
+      request('/profile/' + this.props.match.params.profile_id, "GET")
+      .then(json => {
+          this.setState({ user: json });
+          console.log(json);
```
Merteg: Remove console.log

client/src/components/Profile.jsx
```diff
@@ -1,127 +1,81 @@
...
-        this.setState({ edit: false });
-
-    };
-
-    handleClick_Edit = () => {
-        this.setState({ edit: true });
-    };
+  componentDidMount() {
+      this.setState({ page_id: this.props.match.params.profile_id });
+      request('/profile/' + this.props.match.params.profile_id, "GET")
```
Merteg: `request` has method "GET" by default, you don't need to write it

client/src/components/ProfileEdit.jsx
```diff
@@ -0,0 +1,98 @@
+import React, { Component } from 'react';
+import { Row, Col, Input, Button } from 'react-materialize'
+import { request } from '../utils'
+import { Link } from 'react-router-dom'
+
+
+class ProfileEdit extends Component {
+
+	state = {
+		  user_edit: {},
```
Merteg: Needless tab

client/src/components/ProfileEdit.jsx
```diff
@@ -0,0 +1,98 @@
...
+      error_nick: "",
+      error_repeated: ""
+	}
+
+	componentDidMount() {
+		request('/profile/' + this.props.match.params.profile_id, "GET")
```
Merteg: Bad identation

Merteg (changes requested):

client/src/components/Profile.jsx
```diff
@@ -1,127 +1,81 @@
...
-                        <Row s={1}><h4>{this.state.user.first_name}</h4></Row>
-                        <Row s={1}><h4>{this.state.user.last_name}</h4></Row>
-                        <Row s={1}><h4>{this.state.user.location}</h4></Row>
-                        <Row s={1}><Button waves='light' onClick={this.handleClick_Edit}>Edit</Button></Row>
-                    </Col>
-                </Row>
-                <br />
-              </div>
-            </div>
+  checkID = () => {
```
Merteg: Something like `renderEditButton` seems better

Merteg (commented):
Please make normal indentation, it's really hard to read

yazhyvyi:
I am sorry about inconsistent tabbing in my code, it is Atom editor's issue, resolving it now

yazhyvyi:
Update

client/src/components/ProfileEdit.jsx
```diff
@@ -0,0 +1,148 @@
...
+    render() {
+      return (
+        <div>
+          <div>
+             <h1>Profile</h1>
+          </div>
+          <Row>
+            <Input
+              s={6}
+              error=""
```
RayMet: Why in some inputs do you have empty error or placeholder? Maybe better remove then?

RayMet (changes requested):

client/src/components/ProfileEdit.jsx
```diff
@@ -0,0 +1,148 @@
...
+              id="f_name"
```
RayMet: I think it's better not to reduce the names of id.

RayMet (changes requested):

client/src/components/ProfileEdit.jsx
```diff
@@ -6,93 +6,143 @@ import { Link } from 'react-router-dom'
...
+        request('/profile/' + this.props.match.params.profile_id)
+        .then(json => {
+            this.setState({user_edit: json});
+            this.setState({
+                user_edit: {
+                    f_name: this.state.user_edit.first_name,
+                    l_name: this.state.user_edit.last_name,
+                    nick: this.state.user_edit.nickname,
+                    loc: this.state.user_edit.location,
+                    pass: this.state.user_edit.password
```
Merteg, line 27: I'm really don't understand what is going on here

client/src/components/ProfileEdit.jsx
```diff
@@ -6,93 +6,143 @@ import { Link } from 'react-router-dom'
...
+    }
 
-	UpdateClick = (eve) => {
-		eve.preventDefault()
+    UpdateClick = (eve) => {
+        eve.preventDefault()
 
-    let nick_error = ""
-    let repeated_error = ""
+        let msg_rep = ""
```
Merteg: What `rep` is mean.

client/src/components/ProfileEdit.jsx
```diff
@@ -6,93 +6,143 @@ import { Link } from 'react-router-dom'
...
+                "location": this.state.location,
+                "password": this.state.new_password,
+                "id": this.state.user_edit.id
+            }
+            request('/profile/' + this.props.match.params.profile_id, "PUT", JSON.stringify(data))
+            .then(data => {
+                if (data.msg !== "") {
+                    this.setState({ error_nick: data.msg })
+                }
+                else {
```
Merteg: You are using 'Colander', so if data will be invalid, it returns JSON with `error` field and `if (data.msg !== "")` will return true but you don't handle this error.

Merteg (changes requested):

yazhyvyi:
Update

client/src/components/ProfileEdit.jsx
Merteg: Yes, you say me, that it is `repeat`, so maybe better to name it `msg_repeat`?

client/src/components/ProfileEdit.jsx
```diff
@@ -9,7 +9,12 @@ class ProfileEdit extends Component {
     state = {
         user_edit: {},
         error_nick: "",
-        error_repeated: ""
+        error_repeated: "",
+        error_first: "",
+        error_last: "",
+        error_trigger_nick: false,
+        error_trigger_first: false,
+        error_trigger_last: false
```
Merteg: I don't want to know why did you do this and how it works. But it seems really strange.

client/src/components/ProfileEdit.jsx
Merteg: I think so too. You don't have any reason to type `f_name` instead of `first_name`.

Merteg (commented):
I would not be very happy if I must support this code

server/server/views/user.py
```diff
@@ -0,0 +1,70 @@
...
+
+        response['check_first'] = False
+
+        if data.get('password'):
+            data['password'] = pbkdf2_sha256.hash(data['password'])
+
+        if User.get_one(request, nickname=data.get('nickname')):
+            response['check_nick'] = True
+
+        if data_first is not None:
```
serge30: You have Colander for validating fields, so please use it.

server/server/views/user.py
```diff
@@ -0,0 +1,70 @@
...
+    def put(self):
+        request = self.request
+        data = request.validated
+
+        response = {
+            'check_nick': False,
+            'check_password': False,
+            'check_first': False,
+            'check_last': False
+        }
```
serge30: In case of errors colander provides you `request.errors` you may send back in response and parse it JS to show user.

yazhyvyi:
Update. Fixed merge conflicts

client/src/components/ProfileEdit.jsx
```diff
@@ -0,0 +1,184 @@
...
+                "first_name": this.state.f_name,
+                "last_name": this.state.l_name,
+                "nickname": this.state.nickname,
+                "location": this.state.location,
+                "password": this.state.new_password,
+                "id": this.state.user_edit.id
+            }
+            request('/profile/' + this.props.match.params.profile_id, "PUT", JSON.stringify(data))
+            .then(data => {
+                if (data.check_nick === true) {
```
serge30: It's a bit complicated. You may receive a list of error fields from server along with error messages and apply it form dynamically without hardcoding.

client/src/components/ProfileEdit.jsx
```diff
@@ -36,7 +36,7 @@ class ProfileEdit extends Component {
     UpdateClick = (eve) => {
         eve.preventDefault()
 
-        let msg_rep = ""
+        let msg_for_repeated = ""
```
Merteg: You change it here but you don't change it in the code.

serge30 (changes requested):

Merteg (changes requested):
Also, you don't change the ids of some inputs.

server/server/views/user.py
Merteg: We can't pass a request object to the Colander schema, so one of the possible solutions is to validate everything that requires connection to the database on views

serge30: yep, but looking to code below I do not see any interaction with DB. For access to request object [threadlocal.get_current_request](https://docs.pylonsproject.org/projects/pyramid/en/latest/api/threadlocal.html#pyramid.threadlocal.get_current_request) can be used (it's not recommended to use, but in this case it's acceptable)

yazhyvyi: I have attempted to use Colander and it was partly successful, the issue here was to check if the field `Nickname` is unique, as there is no straightforward method to get the request object, I would need to rely on "hardcoding", which is not acceptable, and I had no idea about `threadlocal.get_corrent_request`. As of now, the code works as intended, and there is little time left, I want to suggest to postpone these requested changes to a separate PR

client/src/components/Profile.jsx
```diff
@@ -1,127 +1,89 @@
...
-            window.Materialize.toast("Profile Updated", 2500);
-          }
-          else {
-            window.Materialize.toast("Incorrect repeated password!", 2500);
-          }
+    renderAvatarImage = () => {
+        if (this.state.user.avatar === null) {
+            return (
+                <img
+                  src="http://dialpharma.com/media/img/default_profile.png"
```
serge30: What is it URL? I cannot access it:
> Not Found
> The requested URL /media/img/default_profile.png was not found on this server.
>
>Additionally, a 404 Not Found error was encountered while trying to use an ErrorDocument to handle the request.

server/server/views/user.py
serge30: We are not in hurry. And it's always a bad idea to merge code which needs to be fixed in future, it needs to be fixed now.

serge30 (changes requested):

client/src/components/Profile.jsx
yazhyvyi: Default profile image, and I can access it on profile page

yazhyvyi:
Update

client/src/components/ProfileEdit.jsx
```diff
@@ -0,0 +1,182 @@
...
+
+    state = {
+        user_edit: {},
+        error_nick: "",
+        error_repeated: "",
+        error_first: "",
+        error_last: "",
+        error_trigger_nick: false,
+        error_trigger_first: false,
+        error_trigger_last: false
```
serge30: Why do you need separate variable for each field, why don't use array or mapping instead?

client/src/components/ProfileEdit.jsx
```diff
@@ -0,0 +1,182 @@
...
+                "password": this.state.new_password,
+                "id": this.state.user_edit.id
+            }
+            request('/profile/' + this.props.match.params.profile_id, "PUT", JSON.stringify(data))
+            .then(data => {
+                try {
+                    if ('errors' in data) {
+                        for (let i = 0; i < data['errors'].length; i++) {
+                            if (data.errors[i].name === "first_name") {
+                                this.setState({ error_first: "First name must not \
```
serge30: You can use error messages received from Colander, instead of hardcoding here

client/src/components/Profile.jsx
serge30: Bu why do you use images from external hosts? Can you be sure that will be there forever? We have local dir to serve a static, please put this file there.

serge30 (changes requested):

client/src/components/ProfileEdit.jsx
```diff
@@ -0,0 +1,182 @@
...
+                    else {
+                        this.setState({ error_nick: "" })
+                        this.setState({ error_repeated: ""})
+                        this.setState({ error_first: ""})
+                        this.setState({ error_last: ""})
+                        this.props.history.push('/profile/' + this.props.match.params.profile_id);
+                        window.Materialize.toast("Profile Updated", 2500);
+                    }
+                }
+                catch (err) {
```
serge30, line 83: What errors are you expecting here?

yazhyvyi:
Update

RayMet (approved):

PR #23, opened 2018-08-21
Feature/map add map

Merteg:
Add map page with map and event markers.


client/src/components/EventForm/index.jsx
```diff
@@ -58,36 +86,54 @@ class AddEvent extends Component {
       start_date: moment(
         [this.state.start_date, this.state.start_time].join(' '),
         'D MMMM, YYYY h:mmA'
-      )._d.toJSON(),
+        )._d.toJSON(),
```
RayMet: I don't understand this "_d.toJSON" why it have "_d."?

RayMet (changes requested):

client/src/components/EventForm/index.jsx
```diff
@@ -115,46 +156,51 @@ class AddEvent extends Component {
             <Input
               s={12}
               id="name"
```
RayMet: I think better change id, It does not describe who have this name.

client/src/components/Nav/NavProfile.jsx
```diff
@@ -17,18 +16,18 @@ class NavProfile extends Component {
...
-                className="user-pictogram"
-                src={this.get_pictogram_url()}
-              />
-              &nbsp;
+            <li>
+              <Link to={"/profile/" + sessionStorage['User-id']}>
+                <img
+                  alt="user pictogram"
+                  className="user-pictogram"
+                  width="32"
```
RayMet: Imagine if it will be open on screen with large resolution, "width=32" will be enough?

RayMet (changes requested):

client/src/components/EventForm/index.jsx
Merteg: moment() return an object, but all I need it is a string with datetime in an appropriate to JSON format.
image

RayMet: But why that have so weird cut name of date?

client/src/components/EventForm/index.jsx
Merteg: It will crash all logic because we have 'name' field in db. input id -> state -> request data -> colander validation -> pyramid view -> db. On all these steps we have field 'name', so change it on one of these steps are not very clever.

client/src/components/EventForm/index.jsx
Merteg: Please ask these question to the moment.js creators, not to me.

client/src/components/EventForm/index.jsx
RayMet: So why I haven't same problems with my id?

client/src/components/Nav/NavProfile.jsx
Merteg: First, it's not part of current PR.
Second, maybe I skip something when I read materializecss source but the biggest size of nav is 64px.
Third, we don't have any validation or ready and working code for images now, so this property is temporary

client/src/components/EventForm/index.jsx
Merteg: If you talk about recover password task because:
first, fields for recover password don't have direct contact with db, but my fields have.
second, we had a conflict with a password for log-in at navbar, so we had to change it

yazhyvyi (commented):
Generally looks good, I tested the code, no errors on runtime. There are some minor things in code i would like to briefly discuss. Any new changes should be applied to the next PR

yazhyvyi (approved):

Merteg:
Resolve conflicts

RayMet (approved):

PR #30, opened 2018-08-27
Feature/edit event page

Merteg:


Merteg:
So many changes, also update db_data_init.py script. Look forward to your review

client/src/components/EventForm/index.jsx
```diff
@@ -86,16 +129,16 @@ class EventForm extends Component {
       start_date: moment(
         [this.state.start_date, this.state.start_time].join(' '),
         'D MMMM, YYYY h:mmA'
-        )._d.toJSON(),
+      ).add(3, 'hours')._d.toJSON(),
```
serge30: You'd better to figure out how to handle time zones in [moment.js](https://momentjs.com/docs/#/manipulating/local/).

client/src/components/EventPage/index.jsx
```diff
@@ -32,18 +39,51 @@ class EventPage extends Component {
...
+        this.setState({
+          category: data.category,
+          tags: data.tags,
+        });
+      };
+    });
+  }
+
+  renderEditButton() {
+    if (Number(sessionStorage.getItem('User-id')) === this.state.author_id) {
```
serge30: To convert string to numbers there is separate function [parseInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt)

server/server/models/event_history.py
```diff
@@ -22,5 +22,15 @@ class EventHistory(Base):
...
         return True
+
+    @classmethod
+    def get_current_event_status(cls, request, event_id):
+        """Get last event history object for some event by id"""
+        max_date_query = request.dbsession.query(func.max(cls.date))\
+            .filter_by(event_id=event_id)
+        return request.dbsession.query(cls)\
+            .filter(cls.event_id == event_id, cls.date == max_date_query)\
+            .first()
```
serge30: You can use just one query to get last event history item:
```
select * 
from event_history
where event_it = :event_id
order by date desc
limit 1
```
Converting this SQL to SQLAlchemy leaving to you :)

server/server/views/event.py
```diff
@@ -91,19 +111,97 @@ def collection_post(self):
...
+            }],
+        }
+        current_event = Event.get_event_by_id(request, self.event_id)
+
+        if current_event is None:
+            raise HTTPNotFound()
+
+        if (current_event.name == request.validated['name'] or
+            Event.get_event_obj(request,
+                                name=request.validated['name']) is None):
```
serge30, line 153: Why do you check presence of `request.validated['name']` in events table? I do not think that having to events with the same name is an error.

server/server/views/event.py
```diff
@@ -91,19 +111,97 @@ def collection_post(self):
...
+                                        currnet_event_tags[tag], self.event_id)
+            for tag in request.validated.get('tags', []):
+                tag_obj = Tag.get_by_name(request, tag.lower())
+                if tag_obj is not None:
+                    tag_id = tag_obj.id
+                    if tag not in currnet_event_tags:
+                        EventTag.add_new(request, tag_id, self.event_id)
+                else:
+                    tag_id = Tag.add_new(request, tag.lower())
+                    EventTag.add_new(request, tag_id, self.event_id)
```
serge30: Block in lines 169-183 is a good candidate to be moved to model. Event to `Event.update_event()` method. It can accept `tags` parameter and update related tables accordingly.

server/server/views/event.py
```diff
@@ -91,19 +111,97 @@ def collection_post(self):
...
+
+            response['success'] = EventHistory.create_new(
```
serge30: You can add method to `Event` model to update state which will create needed `EventHistory` records.

server/server/views/event.py
```diff
@@ -91,19 +111,97 @@ def collection_post(self):
...
+                         "Please wait for review by moderator.")
+            )
+            del response['errors']
+        return response
+
+    @view(permission='update')
+    def delete(self):
+        """Make event closed"""
+        response = {'success': False}
+        response['success'] = EventHistory.create_new(
```
serge30: The same as above,

server/server/views/event.py
Merteg: Because `name` is unique in the database now and I don't sure that the app doesn't crash if I change it :) ... but I will try

server/server/views/event.py
Merteg: I agree with the previous one, but here I don't change any event data, so I don't see the need to move this to the event model.

serge30: The main idea here to hide logic in Model, and controller may even doesn't know anything about EventHistory, as here we deal only with Events.

server/server/views/event.py
serge30: hm... interesting decision :) probably we need to remove this restriction from `Event` model. But it is up to you guys,

Merteg:
Make changes due to the comments.

client/public/index.css
```diff
@@ -62,3 +62,23 @@
...
+    bottom: 70px;
+    padding:0px 0.5rem;
+    text-align: right;
+    background-color: #323232;
+    border-radius: 2px;
+    color: #FFF;
+    width:auto;
+}
+
+blockquote .flow-text {
```
RayMet, line 96: That's doing something?

Merteg: Yes, it removes the margin

client/src/components/EventForm/index.jsx
```diff
@@ -111,23 +161,84 @@ class EventForm extends Component {
...
+      category: this.state.category,
+      start_date: moment(
+        [this.state.start_date, this.state.start_time].join(' '),
+        'D MMMM, YYYY h:mmA'
+      )._d.toJSON(),
+    };
+
+    if (this.state.end_date && this.state.end_time) {
+      eventData.end_date = moment(
+        [this.state.end_date, this.state.end_time].join(' '),
```
RayMet, line 203: This is add finished events to history in database or just confirm finish of the event?

RayMet (changes requested):

client/src/components/EventForm/index.jsx
Merteg: ... sorry but I don't understand.
This part of the code gets date end time from the state (if they exist), converts to JSON valid format and saves in `eventData` object

RayMet (approved):

PR #31, opened 2018-08-28
Feature/main page

Merteg:


yazhyvyi (commented):
When I'm on recover component there is no option to go back to Log-In/Register forms, I suggest you to add "Back" button on recover component for better UX

client/src/components/HomePage/HowAddEvent.jsx
```diff
@@ -0,0 +1,39 @@
...
+import { Row, Col, Card, CardTitle } from 'react-materialize';
+
+
+const HowAddEvent = () => {
+  return (
+    <Row>
+      <Col s={4}>
+        <Card horizontal header={
+          <CardTitle
+            image="http://www.thebluediamondgallery.com/handwriting/images/register.jpg" />
```
serge30: Please do not use links to external resources. Copy them to source tree and use local copies.

client/src/components/HomePage/RecoverPasswordTab.jsx
```diff
@@ -0,0 +1,62 @@
...
+    }
+    if (!emailValidation(this.state.email_recover)) {
+      this.setState({ check_email: 'Invalid' });
+      window.Materialize.toast("Invalid input", 3000);
+      return null;
+    }
+    request('/recover-password', "POST", JSON.stringify(recoverData))
+      .then(data => {
+        this.setState({ msg: data.msg });
+        this.setState({ success: data.success });
```
serge30: You may squash two call to `setState` in one.

client/src/components/HomePage/RecoverPasswordTab.jsx
```diff
@@ -0,0 +1,62 @@
...
+        window.Materialize.toast(this.state.msg, 3000);
+        if (this.state.success) {
+          setTimeout(this.props.history.push('/recover-info'), 3000);
```
serge30: hmm... does it work as expected? As I see `this.props.history.push('/recover-info')` is called immediately. Anyway I'm not good with those delays.

client/src/components/HomePage/Registration.jsx
```diff
@@ -0,0 +1,113 @@
...
+    ) {
+      this.setState({ check_password: "passwords isn't equal" })
+      window.Materialize.toast("Invalid input", 3000)
+      return null
+    }
+    request('/registration', "POST", JSON.stringify(registerData))
+      .then(data => {
+        this.setState({ msg: data.msg })
+        if (
+          this.state.msg === "Your email address is already registered"
```
serge30: Nope. If you need to determine what field is wrong, return this info fro backend instead of comparing error messages, which can be changed in the future.

client/src/components/HomePage/index.jsx
```diff
@@ -0,0 +1,69 @@
...
+            <Route path="/" component={SignInUpTabs} exact />
+            <Route path="/recover" component={RecoverPasswordTab} exact />
+          </Col>
+        </Row>
+      );
+    } else {
+      return (
+        <Row className="img-container">
+          <Col s={10} offset="s1" >
+            <div className="header-text center-align">
```
serge30: There are 2 absolutely identical blocks, it may be a good idea to extract them to separate function/variable?

Merteg:
Update PR

RayMet (approved):

yazhyvyi (approved):
Nice

ykobuk (approved):

PR #32, opened 2018-08-29
Feature/filters on the map

ykobuk:
Add filters by category and days

### DIFF
--- a/client/src/components/MapLayout/index.jsx
+++ b/client/src/components/MapLayout/index.jsx
@@ -9,7 +9,7 @@ import SelectCategory from '../EventForm/SelectCategory';
 class MapLayout extends Component {
   state = {
     events: [],
-    day_filter: 14,
+    day_filter: 0,
     category: ''
   }
 
@@ -24,13 +24,7 @@ class MapLayout extends Component {
       }
       request('/events-short-info', 'POST', JSON.stringify(filterData)).then(data => {
           if ('info' in data) {
-              this.setState({events: []})
-              data.info.forEach(element => {
-                  let eventShortInfo = this.state.events;
-                  eventShortInfo.push(element);
-                  this.setState({ events: eventShortInfo });
-                  console.log()
-              });
+              this.setState({ events: data.info })
           }
       });
   }
@@ -38,16 +32,10 @@ class MapLayout extends Component {
   componentDidMount() {
       let filterData = {
           day_filter: this.state.day_filter
-
       }
       request('/events-short-info', 'POST', JSON.stringify(filterData)).then(data => {
           if ('info' in data) {
-              data.info.forEach(element => {
-                  let eventShortInfo = this.state.events;
-                  eventShortInfo.push(element);
-                  this.setState({ events: eventShortInfo });
-
-              });
+              this.setState({ events: data.info })
           }
       });
   }
@@ -57,9 +45,9 @@ class MapLayout extends Component {
       <Row>
           <Input id='day_filter' s={3} type='select' label="Filter by days" defaultValue='' onChange={this.onChangeHandler}>
               <option value="" disabled selected>Choose filter</option>
-              <option value='1' >today</option>
-              <option value='7'>one weeks</option>
-              <option value='14'>two weeks</option>
+              <option value='1' >Today</option>
+              <option value='7'>One week</option>
+              <option value='14'>Two weeks</option>
           </Input>
           <Button onClick={this.filter}>Filter</Button>
           <Col s={3}>
--- a/client/src/components/MapLayout/index.jsx
+++ b/client/src/components/MapLayout/index.jsx
@@ -1,35 +1,75 @@
 import React, { Component } from 'react';
-import { Row, Col } from 'react-materialize'
+import {Row, Col, Input, Button} from 'react-materialize'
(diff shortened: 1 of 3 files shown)


client/src/components/MapLayout/index.jsx
```diff
@@ -1,35 +1,75 @@
...
+          category: this.state.category
+      }
+      request('/events-short-info', 'POST', JSON.stringify(filterData)).then(data => {
+          if ('info' in data) {
+              this.setState({events: []})
+              data.info.forEach(element => {
+                  let eventShortInfo = this.state.events;
+                  eventShortInfo.push(element);
+                  this.setState({ events: eventShortInfo });
+                  console.log()
```
Merteg: I think you can replace strings 27-32 by `this.setState({ events: data.info })`

client/src/components/MapLayout/index.jsx
```diff
@@ -1,35 +1,75 @@
...
-    request('/events-short-info').then(data => {
-      if ('info' in data) {
-        data.info.forEach(element => {
-          let eventShortInfo = this.state.events;
-          eventShortInfo.push(element);
-          this.setState({ events: eventShortInfo });
-        });
+      let filterData = {
+          day_filter: this.state.day_filter
+
```
Merteg: Useless blank line :)

client/src/components/MapLayout/index.jsx
```diff
@@ -1,35 +1,75 @@
...
       }
-    });
+      request('/events-short-info', 'POST', JSON.stringify(filterData)).then(data => {
+          if ('info' in data) {
+              data.info.forEach(element => {
+                  let eventShortInfo = this.state.events;
+                  eventShortInfo.push(element);
+                  this.setState({ events: eventShortInfo });
+
```
Merteg: The same as on lines 27-32

client/src/components/MapLayout/index.jsx
```diff
@@ -1,35 +1,75 @@
...
+              });
+          }
+      });
   }
 
   render(){
     return (
       <Row>
+          <Input id='day_filter' s={3} type='select' label="Filter by days" defaultValue='' onChange={this.onChangeHandler}>
```
Merteg, line 46: You have 14 days by default in the state, so I think it's better to set it by default here.

server/server/views/location_event.py
```diff
@@ -1,18 +1,33 @@
...
                             cors_origins=('http://localhost:3000',))
 
 
-@events_short_info.get()
+@events_short_info.post()
 def get_events_short_info(request):
     """ """
-    responce = {
+    datetime_now = datetime.now()
+    categories = {'holiday': 1, 'concert': 2, 'movie': 3, 'Voluntary': 4}
```
Merteg: Hardcoding is bad, we have get_all() method in the category model, you can use it.

server/server/views/location_event.py
```diff
@@ -1,18 +1,33 @@
...
-    responce['info'] = Event.get_events_short_info(request, 14)
-    return responce
+    if json['category'] in categories:
+        json['category'] = categories[json['category']]
+
+    if json['category'] == '':
+        response['info'] = Event.get_events_short_info(request, int(json['day_filter']))
+        return response
+    else:
+        response['info'] = request.dbsession.query(Event).\
```
Merteg: Connections to the db must be in models

Merteg (changes requested):

client/src/components/MapLayout/index.jsx
ykobuk: I've left it specially for you =)

ykobuk:
Corrected the problems that Merteg pointed out.

server/server/views/location_event.py
```diff
@@ -11,23 +11,22 @@
...
+    categories_list = [obj.category for obj in categories]
+    categories_id = [obj.id for obj in categories]
+    categories_dict = dict(zip(categories_list, categories_id))
     response = {
         'info': []
     }
-    if json['category'] in categories:
-        json['category'] = categories[json['category']]
+    if json['category'] in categories_list:
+        json['category'] = categories_dict[json['category']]
```
Merteg: I thought a bit, you get a category string from a client, and you need to get it id. So you don't need to get all categories, just get one. (There are a wonderful method in Category model - `get_by_name`)

client/src/components/MapLayout/index.jsx
```diff
@@ -9,7 +9,7 @@ import SelectCategory from '../EventForm/SelectCategory';
 class MapLayout extends Component {
   state = {
     events: [],
-    day_filter: 14,
+    day_filter: 0,
```
Merteg: Why did you remove that? When map renders for the first time, it's don't have any markers now.

Merteg (changes requested):

client/src/components/MapLayout/index.jsx
ykobuk: Because user won't know what is it

Merteg: Hm, I see we must to discuss it

ykobuk:
refactored the location_event.py

Merteg (approved):

PR #33, opened 2018-08-29
Feature/hot events

yazhyvyi:
Hot-Events on Home page


client/public/index.css
```diff
@@ -62,3 +62,23 @@
...
+.hot-event-description {
+    display: block;
+    width: 20em;
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: ellipsis;
+}
+
+.custom-card-height {
+    height: 310px !important;
```
Merteg: Don't use `!important`. It's bad

client/src/components/HotEvents.jsx
```diff
@@ -0,0 +1,68 @@
...
+    })
+  }
+
+  renderHotEvents() {
+      return (
+        <React.Fragment>
+          {this.state.hot_events.map((element) => (
+            <Col key={element.id} s={this.state.display_method}>
+              <Card
+                className={"custom-card-height medium"}
```
Merteg: Why do you use curly braces here?

server/server/views/hot_event.py
```diff
@@ -0,0 +1,46 @@
...
+
+from cornice.resource import resource, view
+from pyramid.security import Allow, Everyone
+
+from ..models import model_to_dict
+from ..models.event import Event
+from ..models.event_history import EventHistory
+
+
+@resource(collection_path='/hot-event/',
```
Merteg: I don't think that the hot events are a resource

server/server/views/hot_event.py
```diff
@@ -0,0 +1,46 @@
...
+
+        histories = EventHistory.get_all_by(request, status_id=5)
+        id_array = [obj.event_id for obj in histories]
+
+        if len(id_array) == 1:
+            response['display_method'] = 12
+        elif len(id_array) == 2:
+            response['display_method'] = 6
+        else:
+            response['display_method'] = 4
```
Merteg: The logic of the server and client must be separate. What happens if you change materialize to something other on the client, you will change this part of the code on the back, it is bad. Better to handle count of the array on the front

server/server/views/hot_event.py
```diff
@@ -0,0 +1,46 @@
...
+    @view(permission='get')
+    def collection_get(self):
+        request = self.request
+
+        response = {
+            'hot_events': None,
+            'display_method': None
+        }
+
+        histories = EventHistory.get_all_by(request, status_id=5)
```
Merteg: You have some problem here. What if event was Hot, but later you change this status to other?
(Record at Event History with status Hot will stay)

Merteg (changes requested):

server/server/views/hot_event.py
yazhyvyi: I think this feature should be implemented on moderator's pages, now you can only set status manually in database

yazhyvyi:
Update

server/server/views/hot_event.py
Merteg: Yes, set status Hot you can only manually, but if the user changes some data in this event, the event status will be "New" with comment - "Wait for review by moderator". All you need it's check the status with the LAST date. Not all event history, just record with the last date

Merteg (approved):

ykobuk (approved):

PR #35, opened 2018-09-02
Fix/sprint3 main page

RayMet:


client/src/components/HomePage/Registration.jsx
```diff
@@ -15,10 +15,41 @@ class Registration extends Component {
...
+    onChangeHandlerPassword = (e) => {
     let { id } = e.currentTarget;
-    this.setState({ [id]: e.currentTarget.value });
-  };
+    this.setState({ [id]: e.currentTarget.value },
+    () => {
+      if (this.state.password === this.state.repeat_password) this.setState({ check_password: "" });
+      else {
+        this.setState({ check_password: "passwords isn't equal" });
+        return null;
```
Merteg: It's a useless line, this function will already end here without it.

client/src/components/HomePage/Registration.jsx
```diff
@@ -15,10 +15,41 @@ class Registration extends Component {
...
+      }
+    });
+    }
+
+    onChangeHandlerEmail = (e) => {
+    let { id } = e.currentTarget;
+    this.setState({ [id]: e.currentTarget.value },
+    () => {
+      if (!emailValidation(this.state.email)) {
+        this.setState({ check_email: "Please provide a valid email address" });
```
Merteg: I know you have a problem with the length of the error message, you can just shorten it to something like "Invalid email"

client/src/components/HomePage/Registration.jsx
```diff
@@ -29,13 +60,17 @@ class Registration extends Component {
...
+      window.Materialize.toast("Invalid input", 3000);
+    }
+    if (!emailValidation(this.state.email)) {
+      this.setState({ check_email: "Please provide a valid email address" });
+    }
+    if (!this.state.nickname.length) {
+      this.setState({ check_nickname: "Empty field" });
       window.Materialize.toast("Invalid input", 3000);
-      return null;
     }
```
Merteg: I think better to stop function if you have some errors before sending a request to the server,

client/src/components/HomePage/Registration.jsx
```diff
@@ -45,10 +80,12 @@ class Registration extends Component {
 
         if ('error' in data) {
           this.setState({ [`check_${data.error}`]: `Invalid ${data.error}` })
-          this.props.history.push('/registration');
           return null;
+
         } else {
-          this.props.history.push('/');
+        window.Materialize.toast("Check your email box", 3000);
+        window.history.forward('/registration-info');
```
Merteg: Wrong indentation. Also, you don't need this `return null`, you don't have any code after this `if`.

client/src/components/HomePage/Registration.jsx
```diff
@@ -57,13 +94,13 @@ class Registration extends Component {
 
   render() {
     return (
-      <div className="white">
+      <div className="white" id="registration-container">
```
Merteg: Why did you add this id?

client/src/components/HomePage/Registration.jsx
Merteg: I don't think it works in right way, better to use `this.props.history.push`

client/src/components/RegistrationInfo.jsx
```diff
@@ -0,0 +1,20 @@
+import React, { Component } from 'react';
+import {Col, Row} from "react-materialize";
+
+class RegistrationInfo extends Component{
```
Merteg: If your component just renders some markup, better to use a function declaration. [Style Guide](https://github.com/airbnb/javascript/tree/master/react#class-vs-reactcreateclass-vs-stateless)

server/server/models/user.py
```diff
@@ -50,6 +50,20 @@ def get_one(cls, request, **kwargs):
...
 
+    @classmethod
+    def get_user_by_email(cls, request, email):
+        """ """
+        return request.dbsession.query(User)\
+            .filter(func.lower(User.email) == func.lower(email)).one_or_none()
+
+    @classmethod
+    def get_user_by_nickname(cls, request, nickname):
+        """ """
```
Merteg: Add a docstrings

server/server/views/auth.py
```diff
@@ -25,7 +26,9 @@ def log_in_post(request):
         'success': False,
         'user': {}
     }
-    user = User.get_one(request, email=request.validated['email'])
+    user = request.dbsession.query(User)\
+        .filter(func.lower(User.email) == func.lower(request.validated['email']))\
+        .one_or_none()
```
Merteg: This does not meet the PEP8 standard. (W503, E128)

Merteg (changes requested):

Merteg:
Also, you have merge conflicts

client/src/components/HomePage/Registration.jsx
RayMet: What is line useless? 22?

client/src/components/HomePage/Registration.jsx
RayMet: Why not? All elements must have their id especially div. But main reason was i tried some css with errors

client/src/components/HomePage/Registration.jsx
RayMet: This.props.history.push do not working. And  window.history.forward works fine, I read a docks and made this lines.

client/src/components/RegistrationInfo.jsx
RayMet: I dont see here problems. Like Serhiy said that must be separate state page.

client/src/components/HomePage/Registration.jsx
Merteg: 25

client/src/components/HomePage/Registration.jsx
Merteg: Now you don't have css for this. `All elements must have their id`, who told you this?

client/src/components/HomePage/Registration.jsx
Merteg: Okay, maybe it's don't work here because of Route in Route, so I don't have any questions for this

client/src/components/RegistrationInfo.jsx
Merteg: You don't understand me, I told about function declaration of a component. Please follow the link

client/src/components/HomePage/Registration.jsx
RayMet: I can delete this. But I dont see a problem for give id. Or just proof that are wrong.

Merteg: The first [link](https://css-tricks.com/the-difference-between-id-and-class/#article-header-id-9) in google. You can find more, but "If you don't need them, don't use them".
The [article](https://dev.to/claireparker/reasons-not-to-use-ids-in-css-4ni4) why `class` is better than `id` in css. If you need css - use class, if you need js - use id.

RayMet:
Update

client/src/components/HomePage/Registration.jsx
Merteg: This can be optimized. You can create bool variable and check it if you have an errors instead of doing validation two times.

client/src/components/HomePage/Registration.jsx
```diff
@@ -81,20 +83,17 @@ class Registration extends Component {
         if ('error' in data) {
           this.setState({ [`check_${data.error}`]: `Invalid ${data.error}` })
           return null;
```
Merteg: I told about all `if` statement, not also about `else`.

server/server/models/user.py
```diff
@@ -52,13 +52,17 @@ def get_one(cls, request, **kwargs):
...
+        """
         return request.dbsession.query(User)\
             .filter(func.lower(User.email) == func.lower(email)).one_or_none()
 
     @classmethod
     def get_user_by_nickname(cls, request, nickname):
-        """ """
+        """Get one user from db by nickname
+        Return user object if user exist and return None if not
+        """
```
Merteg: It does not up to the PEP257 standard. [PEP257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings)

RayMet:
Update, please review.

server/server/models/user.py
Merteg: > Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description.

RayMet:
Update

Merteg (approved):

yazhyvyi (approved):
Nice

PR #36, opened 2018-09-05
Feature/comments mongodb

Merteg:


Merteg:
Update, please review

client/src/components/EventPage/Comments.jsx
```diff
@@ -0,0 +1,163 @@
...
+      .then(data => {
+        if ('success' in data && data.success) {
+          this.getComments();
+
+          let toggleButtons = document.getElementsByClassName('click-to-toggle');
+          for (let i = 0; i < toggleButtons.length; i++) {
+            toggleButtons[i].classList.remove('active');
+          }
+        } else {
+          window.Materialize.toast(`Your comment didn't deleted`, 4000);
```
serge30: 'Your comment hasn't been deleted'

client/src/components/EventPage/Comments.jsx
```diff
@@ -0,0 +1,163 @@
...
+        }
+      })
+  }
+
+  deleteComment = e => {
+    const commentParams = { comment_id: e.currentTarget.getAttribute('data-id') }
+
+    request(`/comment/${this.props.eventId}`, 'DELETE', JSON.stringify(commentParams))
+      .then(data => {
+        if ('success' in data && data.success) {
```
serge30: I think `if (data.success)` is enough

client/src/components/EventPage/Comments.jsx
```diff
@@ -0,0 +1,163 @@
...
+  addComment = e => {
+    let forms = document.getElementsByClassName('answer-form');
+    for (let i = 0; i < forms.length; i++) {
+      forms[i].style.display = 'none';
+    }
+
+    const parentCommentId = e.currentTarget.getAttribute('data-id');
+    let comment = {
+      comment: document.getElementById(`textarea-${parentCommentId}`).value,
+      unix_time: Date.now(),
```
serge30: It's a very bad idea to add timestamps on client side. Never trust data coming from outside, add such info on backend side.

client/src/components/EventPage/Comments.jsx
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  renderChildComments(childs, parentIndex) {
+    if (childs.length) {
+      return (
+        <React.Fragment>
+          {
+            childs.map((obj, index) => {
+              let commentId;
+              if (parentIndex !== '') {
```
serge30: `if (parentIndex) {`

docker-compose.yml
```diff
@@ -8,27 +8,46 @@ services:
...
+            - 5555:5432
+    mongo:
+        image: mongo
+        restart: always
+        container_name: mongo
+        ports:
+            - 27017:27017
+        volumes:
+            - ~/.eventme-mongo:/data/db
+    mongoUser:
```
serge30, line 34: If you need this service only for adding user to connect MongoDB, you need to try using special env variables described in documentation to this Docker image: https://github.com/docker-library/docs/blob/master/mongo/README.md#-via-docker-stack-deploy-or-docker-compose

But as far I remember, MongoDB can be used without any creds defined. You may try this as well.

server/server/models/comment.py
```diff
@@ -0,0 +1,68 @@
...
+            return {'comments': []}
+        return obj['comments']
+
+    @staticmethod
+    def add_comment(request, event_id, author_id, author_nickname, comment_msg,
+                    timestamp, father_comment_id):
+        new_comment = {
+            'author_id': author_id,
+            'author_nickname': author_nickname,
+            'timestamp': timestamp,
```
serge30: It's a good place to generate timestamp, and this should be done using MongoDB functions instead of Python ones.

server/server/models/comment.py
```diff
@@ -0,0 +1,68 @@
...
+            'comment': comment_msg,
+            'child': []
+        }
+        if father_comment_id is None:
```
serge30: Mmmm... I believe `parent` is common name for such things, instead of `father`

server/server/models/comment.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def parse_comment_id(comment_id):
+    """Parse comment_id and return path in db
+
+    Get comment id and return path string (e. g. 'comments.5.child.2.child')
+    Arguments:
+    comment_id -- comment object id (e. g. '5.2')
+    """
+    arrayIndexes = comment_id.split('.')
```
serge30: Mmmmm... What PEP-8 says about style for variable names?

server/server/models/comment.py
```diff
@@ -0,0 +1,68 @@
...
+    path_str = 'comments'
+    for index in arrayIndexes:
+        path_str = path_str + "." + index + ".child"
```
serge30: Something like:
```
path_str = 'comments.' + '.'.join(index + '.child' for index in arrayIndexes)
```

server/server/validation_schema.py
```diff
@@ -69,3 +69,11 @@ class ProfileSchema(colander.MappingSchema):
...
                                    validator=colander.Length(max=255),
                                    missing=colander.drop)
+
+
+class CommentSchema(colander.MappingSchema):
+    """Schema for comments data validation"""
+    comment = colander.SchemaNode(colander.String(),
+                                  validator=colander.Length(max=1500))
+    unix_time = colander.SchemaNode(colander.Int())
+    father_comment_id = colander.SchemaNode(colander.String(), missing=None)
```
serge30: `parent_comment_id`. Please see comment above about naming of this field.

docker-compose.yml
Merteg: I was trying to use env variables, but it's didn't work. Also, I was reading about using MongoDB without credentials, but adminer require the login credentials.

client/src/components/EventPage/Comments.jsx
```diff
@@ -0,0 +1,163 @@
...
+      }
+    })
+  }
+
+  showAnswerForm = e => e.currentTarget.parentNode.parentNode.lastChild.style.display = 'block';
+
+  addComment = e => {
+    let forms = document.getElementsByClassName('answer-form');
+    for (let i = 0; i < forms.length; i++) {
+      forms[i].style.display = 'none';
```
RayMet, line 32: Why it's "none" and not null?

Merteg: because it's css property

server/server/models/comment.py
Merteg: Is it very important to use MongoDB function instead of Python? I need to add a new package for this, but with Python, it's easier.

docker-compose.yml
serge30: That's weird and needs to be investigated. I checked adminer sources and there is case when creds are empty.

server/server/models/comment.py
serge30: Ok, you can use Python's stuff.

docker-compose.yml
Merteg: Okay, I will try again

Merteg:
Update (I could not change docker-compose.yml file)

RayMet (approved):

ykobuk (approved):

PR #37, opened 2018-09-07
Feature/subscribe event

yazhyvyi:
Subscribe feature, also able to view subscribed users (if logged in)


client/.gitignore
```diff
@@ -20,3 +20,4 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 .eslintrc
+package-lock.json
```
Merteg, line 23: I don't think, this file must be here

client/src/components/EventPage/index.jsx
```diff
@@ -52,12 +107,82 @@ class EventPage extends Component {
...
+          })
+        }
+        else {
+          this.setState({
+            is_subbed: false,
+            sub_icon: 'check_circle_outlined',
+            sub_label: 'Subscribe'
+          })
+        }
+        if (data.subscriptions !== null) {
```
Merteg: if (data.subscriptions) seems better

client/src/components/EventPage/index.jsx
```diff
@@ -52,12 +107,82 @@ class EventPage extends Component {
...
+          this.setState({
+            subs: data.subscriptions
+          })
+        }
+        else {
+          this.setState({
+            subs: null
```
Merteg: When you define this state, you write `subs: []`, but here `subs: null`, why?

client/src/components/EventPage/index.jsx
```diff
@@ -20,6 +20,11 @@ class EventPage extends Component {
     long: '',
     category: '',
     tags: {},
+    is_subbed: false,
+    sub_icon: 'check_circle_outlined',
+    sub_label: 'Subscribe',
```
Merteg, line 26: I'm not sure but may be better to leave this states empty by default, you set this data in componentDidMount method.

client/src/components/EventPage/index.jsx
```diff
@@ -36,6 +41,56 @@ class EventPage extends Component {
...
   }
 
+  SubscribeEvent = (eve) => {
+    eve.preventDefault()
+
+    let data;
+    if (sessionStorage['User-id'] !== undefined) {
+      if (this.state.is_subbed === true) {
+        data = {
+          'action': false,
```
Merteg: `action`, what it means? I'm don't understand.

client/src/components/Home.jsx
```diff
@@ -0,0 +1,39 @@
+import React, { Component } from 'react';
```
Merteg: Why did you raise this file from the dead?

Merteg (changes requested):
Hey, this doesn't work, I mean absolutely. You don't have subscribe view.

client/src/components/EventPage/index.jsx
```diff
@@ -36,6 +41,56 @@ class EventPage extends Component {
     }
   }
 
+  SubscribeEvent = (eve) => {
+    eve.preventDefault()
+
+    let data;
+    if (sessionStorage['User-id'] !== undefined) {
+      if (this.state.is_subbed === true) {
```
serge30: Why? What is the reasons  to compare bool variable to `true` or `false`?

yazhyvyi:
Update

server/server/views/subscribe.py
```diff
@@ -0,0 +1,40 @@
+from cornice.resource import Service
+from pyramid.security import Allow, Everyone
+from ..models import model_to_dict
+from ..models.event import Event
+from ..models.subscribe import Subscribe
+from ..models import model_to_dict
+from ..models.event import Event
+from ..models.subscribe import Subscribe
```
Merteg: Duplicate imports

server/server/views/subscribe.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class SubscribeView(object):
```
Merteg: This seems like resource, can you rename it to SubscribeFactory?

server/server/views/subscribe.py
```diff
@@ -0,0 +1,40 @@
...
+     def __init__(self, request):
+         self.request = request
+
+     def __acl__(self):
+         return [(Allow, Everyone, 'get')]
+
+subscribe = Service(name='subscribe', path='/subscribe/{subscribe_user_id}', factory=SubscribeView, cors_origins=('http://localhost:3000',))
```
Merteg: PEP8

Merteg (changes requested):

yazhyvyi:
Update

yazhyvyi:
Update

Merteg (approved):

RayMet (approved):

client/src/components/EventPage/index.jsx
```diff
@@ -53,12 +108,82 @@ class EventPage extends Component {
...
+            is_subbed: true,
+            sub_icon: 'check_circle',
+            sub_label: 'Unsubscribe'
+          })
+        }
+        else {
+          this.setState({
+            is_subbed: false,
+            sub_icon: 'check_circle_outlined',
+            sub_label: 'Subscribe'
```
Merteg: You have this states by default, so you don't need to reset it again.

client/src/components/EventPage/index.jsx
```diff
@@ -53,12 +108,82 @@ class EventPage extends Component {
...
+          })
+        }
+        if (data.subscriptions) {
+          this.setState({
+            subs: data.subscriptions
+          })
+        }
+        else {
+          this.setState({
+            subs: []
```
Merteg: Also, `subs: []` by default

client/src/components/EventPage/index.jsx
```diff
@@ -53,12 +108,82 @@ class EventPage extends Component {
...
       };
     });
   }
 
+  renderSubscribedUsers() {
+    if (this.state.any_subs) {
+      return (
+        <React.Fragment>
+        {this.state.subs.map((element) => {
+          if (element.avatar === null) {
```
Merteg: You duplicate much code here just for an avatar. Better to set it to some variable and just call it later.
Like this: `let avatar = element.avatar === null ? "/person.jpg" : element.avatar` and later ...`src={avatar}`...

server/server/models/subscribe.py
```diff
@@ -17,3 +17,27 @@ class Subscribe(Base):
 
     users = relationship("User", foreign_keys=(user_id,))
     events = relationship("Event", foreign_keys=(event_id,))
+
+    @classmethod
+    def get_subscription(cls, request, **kwargs):
+        subscription = request.dbsession.query(cls).filter_by(**kwargs).\
+                       one_or_none()
+        return subscription
```
Merteg: You don't need to create new variable, just `return request.dbsession.query(cls).filter_by(**kwargs).one_or_none()`

server/server/views/event.py
```diff
@@ -128,8 +130,24 @@ def get(self):
...
-        response['status_str'] = self.event_status
-
+            response['status_str'] = self.event_status
+            subscriptions = Subscribe.get_all_subs(request,
+                            event_id=event_obj.id)
+            subscription = Subscribe.get_subscription(request,
+                           user_id=request.user.id,
+                           event_id=event_obj.id)
+            if subscriptions is not None:
+                id_array = [obj.user_id for obj in subscriptions]
```
Merteg: You can get user objects by the relationships, without id_array.

Merteg (changes requested):
Sorry, I was hurry when approving PR, now I see some problems. Also, check PEP8 in your py files.

yazhyvyi:
Update

Merteg (approved):
Goooood, I think

PR #38, opened 2018-09-07
Feature/fix filter

ykobuk:
filters by drop-down and radio button

### DIFF
--- a/client/src/components/MapLayout/index.jsx
+++ b/client/src/components/MapLayout/index.jsx
@@ -1,80 +1,89 @@
 import React, { Component } from 'react';
-import { Row, Col, Input, Button, Icon } from 'react-materialize'
-import { Link } from 'react-router-dom';
+import { Row, Col, Button} from 'react-materialize'
 import EventMap from '../EventMap';
-import { request, isLogged } from '../../utils';
+import { request } from '../../utils';
 import SelectCategory from '../EventForm/SelectCategory';
 
 
 class MapLayout extends Component {
   state = {
     events: [],
-    day_filter: 0,
+    date_filter: '0',
     category: ''
   }
 
+  onChangeHandlerRadio = (e) => {
+    this.setState({ date_filter: e.currentTarget.value })
+  };
+
   onChangeHandler = (e) => {
-      let {id} = e.currentTarget
-      this.setState({[id]: e.currentTarget.value})
+    let { id } = e.currentTarget
+    this.setState({ [id]: e.currentTarget.value })
+
   };
 
   filter = (e) => {
-      let filterData = {
-          day_filter: this.state.day_filter,
-          category: this.state.category
+    let filterData = {
+      day_filter: this.state.date_filter,
+      category: this.state.category
+    }
+    // alert(filterData.category)
+    request('/events-short-info', 'POST', JSON.stringify(filterData)).then(data => {
+      if ('info' in data) {
+        this.setState({ events: data.info })
       }
-      request('/events-short-info', 'POST', JSON.stringify(filterData)).then(data => {
-          if ('info' in data) {
-              this.setState({ events: data.info })
-          }
-      });
+    });
   }
 
   componentDidMount() {
-      let filterData = {
-          day_filter: this.state.day_filter
-      }
-      request('/events-short-info', 'POST', JSON.stringify(filterData)).then(data => {
-          if ('info' in data) {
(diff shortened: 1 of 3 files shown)


client/src/components/MapLayout/index.jsx
```diff
@@ -1,80 +1,111 @@
...
-
-  renderAddEventButton() {
-    if (isLogged()) {
-      return (
-        <div className="fixed-action-btn horizontal">
-          <Link className="btn-floating btn-large red" to="/add-event">
-            <Icon>add</Icon>
-          </Link>
-        </div>
-      )
```
Merteg, line 52: Turn it back, please. Why did you delete this?

client/src/components/MapLayout/index.jsx
```diff
@@ -1,80 +1,111 @@
...
       }
-      request('/events-short-info', 'POST', JSON.stringify(filterData)).then(data => {
-          if ('info' in data) {
-              this.setState({ events: data.info })
-          }
-      });
+    });
   }
 
   componentDidMount() {
```
Merteg: You have the same parts of code here and in filter method. You can just call a filter method here, I think.

client/src/components/MapLayout/index.jsx
```diff
@@ -1,80 +1,111 @@
...
-              <option value='7'>One week</option>
-              <option value='14'>Two weeks</option>
-          </Input>
+        <Col>
+          <p>
+              <input
+                     className="with-gap"
+                     name="group1"
+                     type="radio"
+                     id="test1"
```
Merteg: :) In my opinion, `test` is not a very suitable id.

client/src/components/MapLayout/index.jsx
```diff
@@ -1,80 +1,111 @@
...
         <Col s={4}>
           <div>
             &nbsp;
           </div>
         </Col>
         <Col s={8} className="map-page-container">
-          <EventMap events={this.state.events}/>
+          <EventMap events={this.state.events} />
         </Col>
-        {this.renderAddEventButton()}
```
Merteg: Also, this must be here.

server/server/views/category.py
```diff
@@ -17,7 +17,7 @@ def __init__(self, request, context=None):
     def __acl__(self):
         return [(Allow, Authenticated, 'get')]
 
-    @view(permission='get')
+    #@view(permission='get')
```
Merteg: Remove...

Merteg (changes requested):

client/src/components/MapLayout/index.jsx
ykobuk: How about id = "radio_button_id_1"?)

Merteg: maybe `radio_1` will be enough. There is no reason to type 'id' in `id` property

ykobuk: May I write radio1 or I must to write underscore between radio and 1?

Merteg: It can be written together, but between different words, you must add an underscore

ykobuk:
deleted some rows and called function filter in componentDidMount

client/src/components/MapLayout/index.jsx
```diff
@@ -9,70 +9,100 @@ import SelectCategory from '../EventForm/SelectCategory';
...
-          </Input>
-          <Button onClick={this.filter}>Filter</Button>
-          <Col s={3}>
-              <SelectCategory onChangeHandler={this.onChangeHandler}/>
-          </Col>
+        <Col>
+          <p>
+            <input
+              className="with-gap"
+              name="group1"
```
RayMet, line 58: I don't like numbers in names, I think better call like group_one or group_one_day, and same with id names.

RayMet (changes requested):

client/src/components/MapLayout/index.jsx
Merteg: Can you argue?

ykobuk: Did you notice that every name='group1' in every input? This is for grouping the radio button, and if i write name="VasiaPypkin" the result will be same! It's just a string that should be the same for all "name" for radio button, it is not even a variable

RayMet (approved):

Merteg (approved):

